PR #16, opened 2018-05-22
Api for all categories and all articles

piyushtiw:


neerajdotname:
@piyushtiw Going forward please do not use the review feature https://how-we-work.bigbinary.com/github/do-not-use-pull-request-reviews-feature.html

config/routes.rb
```diff
@@ -5,4 +5,10 @@
   get "/integrations", to: "application#integrations"
   get "/pricing", to: "application#pricing"
 
+  namespace :api, defaults: {format: 'json'} do
+    namespace :v1, module: 'v1' do
+      get "/all" => "category#all"
```
neerajdotname: @piyushtiw _a 
Later we need to add action for adding, editing and deleting category. So eventually category will turn into a `resource`. If `Category` is a resource then `/all` should be `category#index` so that we are RESTful. 

But the expectation from category#index is that it will return just all categories but here we are returning more.

So Please create a new controller called `Library` and `/all` should go to `library#all`. This will ensure that anytime we are dealing with "library" then we are dealing with both "category" and "articles" together. If we want to purely deal with just "articles" or just "categories" then we can go straight to controller for those.

PR #23, opened 2018-05-24
Auto format elm files on git add

breezyboa:


package.json
```diff
@@ -7,7 +7,19 @@
...
+    "precommit": "lint-staged"
+  },
+  "lint-staged": {
+    "*.elm": [
+      "elm-format --yes",
+      "git add"
+    ]
   }
-}
+}
```
vipulnsward: Missing EOF. @breezyboa can you configure your editor to fix these.

PR #24, opened 2018-05-24
Add loading spinners where applicable

breezyboa:


breezyboa:
@vipulnsward _a

app/javascript/Infix.elm
```diff
@@ -0,0 +1,9 @@
+module Infix exposing (..)
```
vipulnsward: Lets move to Utils module.

Is there no language specific way to achieve this?

app/javascript/Views/Style.elm
```diff
@@ -69,3 +76,17 @@ stylesheet =
...
+    [ Animation.opacity 0
+    , Animation.scale 0.6
+    , Animation.shadow
+        { offsetX = 0
+        , offsetY = 0
+        , size = 20
+        , blur = 0
+        , color = ahGrey
+        }
+    ]
```
vipulnsward: I guess your editor is still not fixed

app/javascript/Views/Style.elm
```diff
@@ -30,6 +30,13 @@ type ArticleStyles
     | Body
 
 
+ahGrey : Color
```
vipulnsward: What does `ahGrey` mean?

app/javascript/Views/Style.elm
```diff
@@ -30,6 +30,13 @@ type ArticleStyles
...
+ahGrey = Color.rgb 153 153 153
+
+
+ahBlue : Color
```
vipulnsward: Same here. I don't understand the name, can you please expand it.

app/assets/stylesheets/spinner.scss
```diff
@@ -0,0 +1,40 @@
...
+      }
+      
+      @keyframes sk-stretchdelay {
+        0%, 40%, 100% { 
+          transform: scaleY(0.4);
+        }  20% { 
+          transform: scaleY(1.0);
+        }
+      }
+}
```
vipulnsward: @rishimohan please take a look at these.

app/javascript/Views/Spinners.elm
```diff
@@ -0,0 +1,21 @@
+module Views.Spinner exposing (..)
+
+import Html exposing (..)
+import Html.Attributes as Attributes exposing (class, style)
+import Color exposing (Color)
+import Color.Convert exposing (colorToCssRgb)
+import Infix exposing ((=>))
+
+
+-- TODO: use style-elements
```
vipulnsward: Lets fix all these in next PR and try to use style-elements package

elm-package.json
```diff
@@ -13,8 +13,10 @@
         "elm-lang/html": "2.0.0 <= v < 3.0.0",
         "elm-lang/http": "1.0.0 <= v < 2.0.0",
         "elm-lang/svg": "2.0.0 <= v < 3.0.0",
+        "eskimoblood/elm-color-extra": "5.1.0 <= v < 6.0.0",
         "mdgriffith/elm-style-animation": "3.5.5 <= v < 4.0.0",
-        "mdgriffith/style-elements": "4.3.0 <= v < 5.0.0"
+        "mdgriffith/style-elements": "4.3.0 <= v < 5.0.0",
+        "rtfeldman/elm-css": "14.0.0 <= v < 15.0.0"
```
vipulnsward: Where/how does elm-css get used?

vipulnsward:
@breezyboa _a added some questions, also needs conflict resolution.

app/javascript/Infix.elm
breezyboa: Did not put this in Utils deliberately since it is recommended that infix operators be separate.
Its better this way for people to search/look for

We could also use (,) a b to create a tuple or the regular (a, b) which is the standard way, but => as infix is pretty common espically when lists of tuples describe lists of key/value pairs like
```
style [ ("color", "#000000"), ("position", "absolute")]
```
this can be written as, Which is a little better readable in my opinion

```
style [ "color" => "#000000",  "position" => "absolute" ]
```

vipulnsward: Ok, this is confusing JS syntax in middle of ELM, lets not do this and stick to ELM way of doing it in a functional way.

breezyboa:
@vipulnsward _a

app/assets/stylesheets/spinner.scss
rishimohan: @vipulnsward Went through the CSS, it seems fine to me.

rishimohan: @breezyboa Hi, instead of using `div` on line 8, we can have a common class for each rectangle, like `rect rect1`, `rect rect2` and then use `rect` instead of div, because div style will be applied to all the divs. Also we follow 2 space indentation so it would be great if that can be followed.

app/assets/stylesheets/spinner.scss
```diff
@@ -0,0 +1,40 @@
...
+    
+          .rect5 {
+            animation-delay: -0.8s;
+          }
+      }
+      
+      @keyframes sk-stretchdelay {
+        0%, 40%, 100% { 
+          transform: scaleY(0.4);
+        }  20% {
```
rishimohan: It's better to separate definitions by lines like for better readability: 
```
0% {
}

100% {
}
```

app/assets/stylesheets/spinner.scss
```diff
@@ -0,0 +1,40 @@
...
+          .rect4 {
+            animation-delay: -0.9s;
+          }
+    
+          .rect5 {
+            animation-delay: -0.8s;
+          }
+      }
+      
+      @keyframes sk-stretchdelay {
```
rishimohan: For cross browser support, we can have same animation with `@-webkit-keyframes`.

PR #62, opened 2018-06-02
Update scss styles

breezyboa:


breezyboa:
@rishimohan _a
have incorporated the review changes you had suggested in the earlier PR

Issue #48, opened 2018-05-29
Add back buttons everywhere


vipulnsward:
https://monosnap.com/file/F7A8zGnpo7Q21PCaIxyHiG1b9fZY2S

https://monosnap.com/file/K3gnAluBH0P0YiMo371pfPp48V4Nq1#

@breezyboa _a

PR #70, opened 2018-06-05
Add a back button

breezyboa:
Addresses #48

### DIFF
--- a/app/javascript/Client/Main.elm
+++ b/app/javascript/Client/Main.elm
@@ -112,7 +112,7 @@ maximizedView : Model -> Html Msg
 maximizedView model =
     let
         history =
-            getFirstHistory model
+            getModelFromHistory model
 
         showBackButton =
             case history of
@@ -218,8 +218,8 @@ transitionFromSection sectionState =
     TransitioningFrom (getSection sectionState)
 
 
-getFirstHistory : Model -> Maybe Model
-getFirstHistory modelHistory =
+getModelFromHistory : Model -> Maybe Model
+getModelFromHistory modelHistory =
     case modelHistory.history of
         ModelHistory model ->
             Just model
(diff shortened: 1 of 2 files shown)


breezyboa:
@vipulnsward _a Please review

app/javascript/Client/Main.elm
```diff
@@ -101,24 +108,44 @@ minimizedView =
...
+        showBackButton =
+            case history of
+                Just previousModel ->
+                    case (getSection previousModel.sectionState) of
+                        Blank ->
+                            False
+
+                        Loading ->
+                            False
+
```
vipulnsward: Aren't these finite?

vipulnsward: I mean the for below line, do we need to catch all by "_"

app/javascript/Client/Main.elm
```diff
@@ -101,24 +108,44 @@ minimizedView =
...
+            (List.concat
+                [ Animation.render model.containerAnimation
+                , [ style
+                        [ ( "position", "fixed" )
+                        , ( "top", "0" )
+                        , ( "background", "#fff" )
+                        , ( "height", "100%" )
+                        , ( "width", "720px" )
+                        , ( "box-shadow", "0 0 50px 0px rgb(153, 153, 153)" )
+                        , ( "font-family", "proxima-nova, Arial, sans-serif" )
```
vipulnsward: Hmm, aren't we we moving to styled css yet?

breezyboa: @vipulnsward no at least not yet. But I will be moving all such styles to scss files.
The current state of styled elements have certain limitations that are a bit tricky to get around. It makes the code more convoluted. Might as well move them all out to separate files. I will be doing this in a separate PR in which I will also improve the ui

app/javascript/Client/Main.elm
breezyboa: Did not get you.

vipulnsward:
Please rebase

PR #80, opened 2018-06-08
73 cors issue

piyushtiw:


piyushtiw:
@vipulnsward please review it.

config/initializers/cors.rb
```diff
@@ -3,7 +3,7 @@
 # enable cors
 Rails.application.config.middleware.insert_before 0, Rack::Cors do
   allow do
-    origins '*\.(heroku|acehelp)\.com'
+    origins '.*(heroku|acehelp).com'
```
rkr090: - The regex should be strict, this will also allow ```*.acehelp.com.attacker.co.in```. How about ```.*(heroku|acehelp).com\z```
- If the only allowed schemes are **http** and **https**, why not use: ```https?:\/\/.*(heroku|acehelp).com\z```
- Let's not allow heroku.com, just allow herukoapp.com, ```https?:\/\/.*(herokuapp|acehelp).com\z```
- The above regex also allows **https://staging.hahad-acehelp.com/**, lets use ```https?:\/\/([a-z0-9-]+[.])?(herokuapp|acehelp).com\z```

```https?:\/\/([a-z0-9-]+[.])?(herokuapp|acehelp).com\z```

Note 1: The regex does not allow trailing '/' character.
Note 2: One issue with the above regex is that it also allows domain/subdomains starting with character '-' which is not valid. We can fix it at the expense of complicating the regex using lookbehinds.

@piyushtiw _a You can test the regex on http://rubular.com/
@vipulnsward Let me know your thoughts, I just want to be slightly strict with which domains we allow. It's a low hanging security issue.

piyushtiw:
@rkr090 _a please review this.

PR #93, opened 2018-06-15
cors issue for herokuapp

piyushtiw:


piyushtiw:
@vipulnsward _a please review it.

config/initializers/cors.rb
```diff
@@ -3,12 +3,7 @@
...
 Rails.application.config.middleware.insert_before 0, Rack::Cors do
   allow do
-    origins 'http://staging.acehelp.com'
-    resource "*", headers: :any, methods: [:get, :post, :put, :delete, :options]
-  end
-
-  allow do
-    origins 'https://staging.acehelp.com'
-    resource "*", headers: :any, methods: [:get, :post, :put, :delete, :options]
+    origins /\Ahttps?:\/\/.*(herokuapp|acehelp).com\z/
```
vipulnsward: Lets do ENV wise. 

On production only allow acehelp.com .

On staging allow herokuapp.

 On development, nothing should prevent this.

vipulnsward:
please assign to Rohit after changes

piyushtiw:
@rkr090 _a please review it.

config/secrets.yml
```diff
@@ -45,3 +48,4 @@ production:
     mailer_sender: "'Learn Academy' <do-not-reply@acehelp.com>"
     secret_key: <%= ENV['DEVISE_KEY'] %>
   api_base_url: ''
+  allowed_origins: 'https://acehelp.com'
```
rkr090: only https?

config/secrets.yml
```diff
@@ -37,6 +39,7 @@ staging:
     mailer_sender: "'Learn Academy' <do-not-reply@acehelp.com>"
     secret_key: 0d1067153029e34a519dcf73db39b77504b7c3c856673ff0f361d9c5eeca1cfd28d639692d55d35dda28bc265195a87f3b3e42f4947c8f4c144723a8f0866120
   api_base_url: 'http://staging.acehelp.com/api/v1'
+  allowed_origins: /\Ahttps?:\/\/.*(herokuapp|acehelp).com\z/
```
rkr090: @piyushtiw This is not working, check https://acehelp-staging-pr-93.herokuapp.com/
The regex in #80 also didn't work

config/initializers/cors.rb
```diff
@@ -1,14 +1,10 @@
...
-Rails.application.config.middleware.insert_before 0, Rack::Cors do
-  allow do
-    origins "http://staging.acehelp.com"
-    resource "*", headers: :any, methods: [:get, :post, :put, :delete, :options]
-  end
-
-  allow do
-    origins "https://staging.acehelp.com"
-    resource "*", headers: :any, methods: [:get, :post, :put, :delete, :options]
+module Api
```
rkr090: The module name ```Api``` is incorrect, it should be ```Acehelp```. The second thing is we don't need to wrap it inside the module.

config/secrets.yml
piyushtiw: should I allow http on prod?

config/secrets.yml
piyushtiw: made some changes, This time I am not giving this regex as a string.

rkr090: The new and the old regex both matches the URL ```\Ahttps?:\/\/.*(herokuapp|acehelp).com\z```.

But if you check https://acehelp-staging-pr-93.herokuapp.com/ it is not working.

config/secrets.yml
rkr090: We don't have production right now. I assume we will force redirect HTTP to HTTPS cc: @vipulnsward ?

vipulnsward:
@rkr090 so what is the resolution here, we don't need cors right now right? 

Where is the fix actually needed?

vipulnsward:
Closing in favour of https://github.com/bigbinary/acehelp/issues/97

PR #99, opened 2018-06-16
Graphql optimization

piyushtiw:


piyushtiw:
@rkr090 _a I have changed all the N+1 queries in Graphql in a single query. Please review this.

PR #130, opened 2018-06-25
suggested page no url match

piyushtiw:


piyushtiw:
@vipulnsward _a

vipulnsward:
No, we should instead ask users to "visit" all articles?

@breezyboa _a probably something to be handled on ELM side?

breezyboa:
@vipulnsward can be handled in Elm.
Do you think it will be better to Auto redirect to All if we get a 400 bad request on suggested articles?

vipulnsward:
Hmm, we can't do that, we always have the suggested articles tab.
Lets implement showing user a "No suggested article" or something and guiding them to look at "All articles" page

We can discuss more what feels better after this is implemented.

vipulnsward:
@breezyboa _a

breezyboa:
Assigning it to @piyushtiw _a

piyushtiw:
refer to https://github.com/bigbinary/acehelp/pull/144

PR #131, opened 2018-06-25
Fix broken admin pages on staging

### LINKED ISSUE #121: Admin pages are broken on staging
https://monosnap.com/file/FCvbnhUrLT9FjmDhy0ZaIkUIdAzKly

@sudeeptarlekar _a

sudeeptarlekar:
This fixes #121 

Admin pages are broken on staging because for time being I have hard-coded organization API keys in code and those keys were not present on staging.

Added rails code which will fetch first organization key(again when a user to organization mapping is finalized we will have to fetch user specific organization keys) and pass it to Elm frontend which will use passed key in an API call.

@breezyboa _a please review


breezyboa:
@sudeeptarlekar _a 
please fix the build

app/views/admin/urls/new.html.erb
```diff
@@ -1,2 +1,4 @@
 <%= stylesheet_pack_tag 'admin' %>
 <%= javascript_pack_tag "admin" %>
+<div id="organization_key" data-organization-key="<%= @organization.api_key %>">
+<div id="url" data-url="<%= @url.url%>" >
```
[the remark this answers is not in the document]
sudeeptarlekar: so an url is being used to fetch a list of articles associated with an organization. It is being passed as a param to an API.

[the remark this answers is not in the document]
breezyboa: Is this for article listing? won't the default page list out all available articles in the organization? 
Also even if the user had to select a URL he/she should be able to select that particular URL from drop down or something. Correct me if I am wrong

sudeeptarlekar:
@breezyboa _a fixed build

app/controllers/concerns/set_organization.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+# This temporary, remove this concern once mapping between user and organization is created
+module SetOrganization
```
vipulnsward: Instead of this concern, why not add include this in application controller, and then only add a before action wherever needed?

app/controllers/concerns/set_url.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+# This is temporary, remove this concern once mapping between user and organization is created
+module SetUrl
```
vipulnsward: Ditto

app/controllers/concerns/set_organization.rb
sudeeptarlekar: yes, that's correct, but we are following this practice in Aceinvoice-Web where action declaration and usage is at the same level, so I thought let's follow same here

vipulnsward: Can we get rid of this complexity?

vipulnsward:
@sudeeptarlekar can we please merge this today?

breezyboa:
@sudeeptarlekar  _a

sudeeptarlekar:
@breezyboa _a

PR #132, opened 2018-06-25
Conventions for GraphQL errors

abdul-shajin:
error messages for absence of api-key
unauthorized request
preventing DDOS attack by limiting the depth to 8
Handling Mutation Errors with a new type Error
test coverage for error responses
abstraction function for render unauthorized action
load_organization: error handling made better


vipulnsward:
@abdul-shajin what is the status on this?  Is this nearing completion?

vipulnsward:
@abdul-shajin _a

app/controllers/concerns/load_organization.rb
```diff
@@ -11,9 +11,37 @@ def load_organization
     if Rails.env.development?
       @organization = Organization.first
     else
-      @organization = Organization.find_by(api_key: request.headers["api-key"])
+      api_key = request.headers["api-key"]
+
+      render_unauthorized_request(err: :missing_keys) && return if api_key.blank?
+      @organization = Organization.find_by(api_key: api_key)
     end
+    api_key = request.headers["api-key"]
```
piyushtiw: This should be in else block.

app/controllers/graphql_controller.rb
```diff
@@ -7,7 +7,9 @@ def execute
     result = AcehelpSchema.execute(query, variables: variables, context: context, operation_name: operation_name)
     render json: result
   rescue => e
-    render json: { error:  e.message }, status: 500
+    show_error_in_development(e)
```
piyushtiw: Need not to show error explicitly in development.

app/graphql/mutations/article_mutations.rb
```diff
@@ -9,20 +9,27 @@ class Mutations::ArticleMutations
...
-        { article: new_article }
+      if category.nil?
+        errors = Utils::ErrorHandler.generate_error_hash('Category not found', context)
       else
-        GraphQL::ExecutionError.new("Invalid input: #{new_article.errors.full_messages.join(', ')}")
+        new_article = category.articles.new(title: inputs[:title], desc: inputs[:desc])
+        new_article.organization = context[:organization]
+
+        if new_article.save
+          article = new_article
```
piyushtiw: extra variable

app/graphql/mutations/article_mutations.rb
```diff
@@ -39,17 +46,25 @@ class Mutations::ArticleMutations
...
-      return { errors: "Article not found" } if article.nil?
 
-      if article.update_attributes(inputs[:article].to_h)
-        { article: article }
+      if article.nil?
+        errors = Utils::ErrorHandler.generate_error_hash('Article not found', context)
       else
-        GraphQL::ExecutionError.new("Invalid input: #{article.errors.full_messages.join(', ')}")
+        if article.update_attributes(inputs[:article].to_h)
+          updated_article = article
```
piyushtiw: ditto

app/controllers/graphql_controller.rb
abdul-shajin: When I watch development log, This wasn't make much sense to see no info about the error occurred.
Will remove it anyway

app/graphql/mutations/article_mutations.rb
abdul-shajin: Actually not, When api encounters any errors, Along with errors field article fields should also being filled-in even though  its not saved in database. Do you really think it has to be that way?
`{ article: null, errors: [ {path: "a"} ]`  will make more sense compared to 

`{ article: {id: null, name: 'Somename'}, errors: [ {path: "a"} ]`

vipulnsward:
@abdul-shajin when you are done assign it to @sudeeptarlekar for review

app/controllers/graphql_controller.rb
neerajdotname: I disagree with Piyush strongly. Why would you not want to see error in development. @piyushtiw can you explain.

app/graphql/mutations/article_mutations.rb
neerajdotname: All comments regarding extra variable are not valid. It seemed @piyushtiw thoughts some variables were unused which is not true.

piyushtiw: We are getting errors as follow at present
```
mutation addContact{
  addContact(input: {name: "", email: "dummy@email.com", message: "dummy"}){
    contact{
      id
      name
    }
  }
}
```

```
{
  "data": {
    "addContact": null
  },
  "errors": [
    {
      "message": "Invalid input: Name can't be blank",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "addContact"
      ]
    }
  ]
}
```



But when we add extra variable in response with the same query above it is giving

```
{
  "data": {
    "addContact": {
      "contact": null
    }
  }
}
```

In this case we have to change the query explicitly and get error in response like below

```
mutation addContact{
  addContact(input: {name: "", email: "dummy@email.com", message: "dummy"}){
    contact{
      id
      name
    }
    errors {
      path
      message
    }
  }
}
```

it is giving :-

```
{
  "data": {
    "addContact": {
      "contact": null,
      "errors": [
        {
          "path": [
            "addContact",
            "attribute",
            "name"
          ],
          "message": "name can't be blank"
        }
      ]
    }
  }
}
```

app/controllers/graphql_controller.rb
piyushtiw: Yeah sorry, my bad. The reason behind this was, we would get the error in response if we would not log the error explicitly in development also. In future, if we would have to debug the code, we can change these lines.

neerajdotname: > we would get the error in response if we would not log the error explicitly in development also

I'm not following you. Whether we log something or not - how does that impact what is in response. What am I missing.

app/graphql/mutations/article_mutations.rb
neerajdotname: Thanks for the detailed explanation. However because of my lack of knowledge in GraphQL I'm not able to follow it. Please ping me on skype when you can have a call.

app/controllers/graphql_controller.rb
piyushtiw: Please go through it http://graphql-ruby.org/queries/backtrace_annotations.html#wrapping-errors

app/controllers/concerns/load_organization.rb
```diff
@@ -11,9 +11,36 @@ def load_organization
...
+      render_unauthorized_request(err: :missing_keys) && return if api_key.blank?
+      @organization = Organization.find_by(api_key: api_key)
     end
 
-    render_unauthorized("Unauthorized request") && return unless @organization
+    render_unauthorized_request(err: :missing_keys) && return if api_key.blank?
+    @organization = Organization.find_by(api_key: api_key)
+    render_unauthorized_request && return if @organization.blank?
+    
+    @organization.inspect
```
vipulnsward: Why the inspect?

app/controllers/concerns/load_organization.rb
```diff
@@ -11,9 +11,36 @@ def load_organization
...
+  end
+
+  private
+  def graphql_api_call?
```
vipulnsward: Add a line before the method

Why is this not indented correctly? Dint run rubocop on precommit?

app/controllers/concerns/load_organization.rb
abdul-shajin: inspect was there in master branch, but not now. It seems while resolving the conflict, I reverted back the removal. My bad. Will remove the line.

app/controllers/concerns/load_organization.rb
abdul-shajin: Never used to add indentation after access modifiers. Read that somewhere long back that its not a good practice.
But yes, will make it a habit following rubocop suggestions and add it in pre-commit hook.

app/graphql/mutations/article_mutations.rb
```diff
@@ -9,19 +9,20 @@ class Mutations::ArticleMutations
     input_field :desc, !types.String
 
     return_field :article, Types::ArticleType
-    return_field :errors, types.String
```
vipulnsward: Is this the preferred way? I am not sure why we return a single string or all fields errors. 
They should be returned as an array/hash of errors individually.

cc @rkr090

PR #144, opened 2018-06-27
text msg for no suggested articles

piyushtiw:
@breezyboa _a please review this.


breezyboa:
@piyushtiw have updated the message and presented it in a better way

Whenever we take up an issue lets start thinking and solving it in terms of how will the end user be experiencing the solution. In this case, we don't simply want the text "OOPS! No suggested articles!" to show up if there are no suggested articles. What is the user supposed to do next? 

The changes that Ive pushed are also not exactly accurate but it is still better than showing a plain text message and directs the user to take the next action i.e to visit the Library tab or to search for an article

cc: @sudeeptarlekar

PR #153, opened 2018-06-28
Add JS snippet to be used by client websites

breezyboa:
The template is right now in a js file. This should be probably moved to a proper template later on

Also since webpacker adds a hash value to each public asset and since with each deployment or build the filename for the client javascript and css files will change (if there any changes to those files), we use the manifest.json file which is also generated by webpack to refer to the actual generated file names. 
This is also probably not the best way to do it and should be updated later.


breezyboa:
@vipulnsward  _a can you review this

PR #166, opened 2018-06-30
serach articles ids in int

piyushtiw:


piyushtiw:
@vipulnsward _a please review it.

vipulnsward:
@breezyboa _a I cleaned up the backend, but this is actually decoded issue, why is the decoder assuming the id should be an integer?

Issue #167, opened 2018-07-02
Elm - UI for "Did this answer your question" feature


breezyboa:
Parent issue: #147 

Refer:
https://whimsical.co/Sx47eHTG7FAQxng15JS9F1

PR #176, opened 2018-07-02
"Did this article help" feature

breezyboa:
Addresses #167


breezyboa:
On hold till backend work is completed

neerajdotname:
@breezyboa _a what do you need from the backend. I can take a look at it.

breezyboa:
@neerajdotname _a 
For the full feature to be complete #180 needs to be worked on

neerajdotname:
@breezyboa _a Can you wire up the HTTP call for upvote. If that works then I'll work on downvotes and tests.

neerajdotname:
Added support for registering downvote.

breezyboa:
@neerajdotname  added calls for both upvote and downvote from elm side

neerajdotname:
@breezyboa if I hit "downvote" and then submit the form with just filling "comment" then I'm getting "something went wrong" error. Is this working for you ?

breezyboa:
@neerajdotname Form submission is not yet complete so its always showing "something went wrong". Will add that next once the grahpQl mutation is ready

neerajdotname:
@vipulnsward _a Please review it. I have not added tests for mutations. Soon we will write cypress test for all cases so we can go a bit easy on tests for now.

app/graphql/mutations/article_mutations.rb
```diff
@@ -94,4 +94,86 @@ class Mutations::ArticleMutations
...
+    }
+  end
+
+  Upvote = GraphQL::Relay::Mutation.define do
+    name "Upvote"
+
+    input_field :id, !types.ID
+
+    return_field :article, Types::ArticleType
+    return_field :errors, types.String
```
vipulnsward: `return_field :errors, types[Types::ErrorType]`

vipulnsward: Same at other places

app/javascript/Client/Section/Article.elm
```diff
@@ -1,39 +1,178 @@
...
+httpVote : Encode.Value -> Decode.Decoder String -> NodeEnv -> Cmd Msg
+httpVote encode decode env =
+    Http.send Vote <| Http.post (graphqlUrl env) (Http.jsonBody encode) decode
+
+
+update : NodeEnv -> Msg -> Model -> ( Model, Cmd Msg )
+update env msg model =
+    case msg of
+        FeedbackSelected feedback ->
+            -- TODO: Make and Http call to register feedback
```
vipulnsward: Is this todo still valid?

app/models/article.rb
```diff
@@ -11,4 +11,13 @@ class Article < ApplicationRecord
   validates :title, uniqueness: { scope: [:organization_id, :category_id] }, presence: true
 
   scope :for_organization, ->(org) { where(organization: org) }
+
+  def increment_upvote
+    self.update(upvotes_count: self.upvotes_count + 1)
```
vipulnsward: Can use http://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-increment and save

vipulnsward: nvm

db/migrate/20180704122137_add_downvotes_to_articles.rb
```diff
@@ -0,0 +1,5 @@
+class AddDownvotesToArticles < ActiveRecord::Migration[5.2]
+  def change
+    add_column :articles, :downvotes_count, :int, default: 0
```
vipulnsward: Lets club 2 Migrations into 1, they are not doing anything specific for them to be in 2

app/assets/stylesheets/client/article.scss
```diff
@@ -0,0 +1,83 @@
...
+        }
+        .thumbs-up {
+            margin-top: -4px;
+
+            &.selected {
+                background-color: #eb6464
+            }
+        }
+    }
+}
```
vipulnsward: @rishimohan to check the CSS changes

app/graphql/mutations/article_mutations.rb
```diff
@@ -94,4 +94,86 @@ class Mutations::ArticleMutations
...
+    name "Downvote"
+
+    input_field :id, !types.ID
+
+    return_field :article, Types::ArticleType
+    return_field :errors, types.String
+
+    resolve ->(object, inputs, context) {
+      article = Article.find_by(id: inputs[:id], organization_id: context[:organization].id)
+      return { errors: "Article not found" } if article.nil?
```
vipulnsward: Please see https://github.com/bigbinary/acehelp/pull/132/files on how we handle the missing object errors now.

I will cleanup more on that later, but lets use that on so I can cleanup as one single change.

app/graphql/mutations/article_mutations.rb
```diff
@@ -94,4 +94,86 @@ class Mutations::ArticleMutations
...
+
+      if article.increment_downvote
+        { article: article }
+      else
+        GraphQL::ExecutionError.new("Invalid input: #{article.errors.full_messages.join(', ')}")
```
vipulnsward: ditto

vipulnsward:
@neerajdotname _a added some comments, can merge after that.

cc @breezyboa as well

db/migrate/20180704122137_add_downvotes_to_articles.rb
neerajdotname: I disagree on this. This is causing more work without any benefit.

neerajdotname:
@abdul-shajin _a Can you take care of the feedback given by Vipul for me. I won't be able to get to work on these today and I do not want this PR to hold other work or delay it to the point it might cause merge conflicts.

abdul-shajin:
Yes, will fix it before EOD

breezyboa:
Have added in Article feedback form submission through `addContact` mutation
Note that this mutation requires all fields be non-empty. Whereas the actual requirement is that only the comment/message is the mandatory field

app/assets/stylesheets/client/article.scss
```diff
@@ -0,0 +1,83 @@
...
+        textarea, input[type=text] {
+            width: 100%;
+        }
+    }
+
+    .did-this-help{
+        display: flex;
+        justify-content: center;
+        
+        div{
```
rishimohan: Let's use class for elements instead of using the tag, reason being if we add another `div` then the CSS will apply to that `div` too. Also if there's any child div of this div, then style will apply to that too. Better to keep the specificity, same with the span definition below. 

Also let's keep consistency and add a space between class/id name and the opening bracket. Ex: `div {`, `.did-this-help {`

breezyboa:
@sudeeptarlekar  _a

PR #182, opened 2018-07-04
Fix breaking input box styles

breezyboa:


breezyboa:
Merging this as its a small style fix

PR #185, opened 2018-07-05
Added User & Organization Mutation

abdul-shajin:
Issue# 151
Added User Mutation
Added Organization Mutation
Incorporated new error handling
Added new migration(organization_id in users, email in organizations)
Errors should show in logs irrespective of the environment we're in
Raising the exception if its in test as its difficult to understand what went wrong with current approach
New UserType(GraphQL)
Organization-User association changes, (org has many users)
Rubocop Suggestions


sudeeptarlekar:
Looks good to me, but as I don't have much knowledge of GraphQL, I request you @neerajdotname _a to take a look and merge it.

Thanks in advance

app/graphql/mutations/organization_mutations.rb
```diff
@@ -0,0 +1,34 @@
...
+    input_field :user_id, !types.ID
+    input_field :name, !types.String
+    input_field :email, !types.String
+
+    return_field :organization, Types::OrganizationType
+    return_field :errors, types[Types::ErrorType]
+
+    resolve -> (object, inputs, context) {
+      user = User.find_by_id(inputs[:user_id])
+      if !user
```
neerajdotname: @abdul-shajin Since you are handling both the conditions do not use negation condition. It is much easier to understand following.

```
If user
   x
else
   y
end
``

It is harder to understand this.

```
If !user
   y
else
   x
end
```

File: app/graphql/mutations/organization_mutations.rb [hunk 2]
```diff
@@ -0,0 +1,34 @@
...
+        errors = Utils::ErrorHandler.new.generate_error_hash("User not found", context)
+      else
+        valid_params = inputs.to_h.slice("name", "email", "user_id")
```
neerajdotname: `valid_params` works. However I prefer `sanitized_params`. Word `sanitized` adds a different value and in this case I think we are sanitizing the list and thus the data set becomes valid also.

app/graphql/mutations/user_mutations.rb
```diff
@@ -0,0 +1,31 @@
...
+    input_field :first_name, !types.String
+    input_field :last_name, types.String
+    input_field :password, !types.String
+    input_field :password_confirmation, !types.String
+
+    return_field :user, Types::UserType
+    return_field :errors, types[Types::ErrorType]
+
+    resolve ->(object, inputs, context) {
+      new_user = User.new(inputs.to_h)
```
neerajdotname: @abdul-shajin Please sanitize the list. If we do not then user can make onself an admin by passing `admin: true`. There is no `admin` role right now but the point is that we need let only certain attributes come through.

app/models/user.rb
```diff
@@ -5,4 +5,18 @@ class User < ApplicationRecord
   # :confirmable, :lockable, :timeoutable and :omniauthable
   devise :database_authenticatable, :registerable,
          :recoverable, :rememberable, :trackable, :validatable
+
+  validates :first_name, presence: true
+
+  # Associations
```
neerajdotname: Please remove this comment. It does not add much value.

config/routes.rb
```diff
@@ -23,6 +23,13 @@
       get "/articles/search", to: "article#search"
 
       resource :contacts, only: :create
+
+      get "/organization/:organization_id/data", to: "organization#data"
```
neerajdotname: @abdul-shajin Is there a reason why these routes are needed? Are we not doing graphql for these actions?

db/migrate/20180629152054_add_email_to_organizations_and_organization_id_to_users.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddEmailToOrganizationsAndOrganizationIdToUsers < ActiveRecord::Migration[5.2]
+  def change
+    add_column :organizations, :email, :string
```
neerajdotname: Since email is required let's make it not nullable.

db/migrate/20180629152054_add_email_to_organizations_and_organization_id_to_users.rb
```diff
@@ -0,0 +1,8 @@
...
+    add_column :users, :organization_id, :string
```
neerajdotname: Please add a foreign key for `organization_id`.

sudeeptarlekar:
@vipulnsward _a PR is ready to review

app/javascript/Admin/Main.elm
```diff
@@ -347,19 +347,19 @@ retriveOrganizationFromUrl : Location -> OrganizationId
...
-getOrganizationId : Maybe Int -> OrganizationId
+getOrganizationId : Maybe String -> OrganizationId
 getOrganizationId orgId =
     case orgId of
         Just orgId ->
             orgId
 
         Nothing ->
-            -1
+            "abcd"
```
vipulnsward: ""

app/javascript/Admin/Page/Url/List.elm
```diff
@@ -22,7 +22,7 @@ type alias Model =
 initModel : Model
 initModel =
     { listOfUrls = { urls = [] }
-    , urlId = 0
+    , urlId = "abcd"
```
vipulnsward: same

app/views/devise/confirmations/new.html.erb
```diff
@@ -0,0 +1,16 @@
+<h2>Resend confirmation instructions</h2>
```
vipulnsward: Convert to use slim.

vipulnsward:
@breezyboa _a Please review ELM side.

breezyboa:
@vipulnsward _a
Elm code looks good

vipulnsward:
@prathamesh-sonpatki _a

Issue #124, opened 2018-06-21
Html elements with data-acehelp-article should open acehelp widget with the article id


breezyboa:
eg: 
```html
<a href="#" data-acehelp-article="1234" data-acehelp-style="underline">Open Article</a>
```
should open widget with article id 123
This should be applicable for any html element

PR #188, opened 2018-07-09
Bind html elements with data-acehelp-article attributes

breezyboa:
Addresses #124 

An example of using this feature:
```html
<a href="#" data-acehelp-article="2">Open Article with id:2</a>
```


breezyboa:
@sudeeptarlekar _a can you review this

PR #192, opened 2018-07-12
change addContact to addTicket

neerajdotname:


neerajdotname:
@breezyboa _a

neerajdotname:
@sudeeptarlekar I'm trying to make `name` optional.

However when I do not pass `name` or pass `name` as empty string then ticket record is not being recorded. Do you have any idea what the fix is https://monosnap.com/file/lviUruKIb9Amyy8rOiPUnWTBmuRBTy

breezyboa:
@neerajdotname since tickets are created against an article. Shouldn't the addTicket mutation take in an article id also as an argument to be associated with an article? 
Also currently in this mutation, the email field is mandatory. According to the requirement, the email field should be optional.

neerajdotname:
You are right. It should take article_id.

Email is required for a ticket. If email is not filled then client needs to call addFeedback and not addTicket.

breezyboa:
changed from addContacts to addTicket in elm
Note that name is also considered as a mandatory field. This should be an optional field in this mutation

once addFeedback mutation is ready, will use it when email is not provided

breezyboa:
@neerajdotname _a

neerajdotname:
@breezyboa probably you missed this comment of mine https://github.com/bigbinary/acehelp/pull/192#issuecomment-404559188

breezyboa:
Oh. yes sorry my bad

vipulnsward:
@breezyboa _a All this is still using old code and old-routes: https://monosnap.com/file/v14GAlJgeeRAaRLFgYqNdbP5ZoD1Rb

Please fix the requests on this page as well as the names

breezyboa:
@vipulnsward _a 
changed from contact api to addTicket graphql mutation in contact us screen

app/controllers/api/v1/ticket_controller.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class TicketController < BaseController
```
vipulnsward: @sudeeptarlekar _a can you check if this is redundant now?

vipulnsward:
@prathamesh-sonpatki _a

app/javascript/Client/Request/Article.elm
```diff
@@ -41,7 +42,13 @@ requestSearchArticles searchTerm =
...
 
 
--- GRAPHQL
+-- requestFeedbackMutation : FeedbackForm -> Reader ( NodeEnv, ApiKey ) (Task GQLClient.Error (Maybe (List GQLError)))
+-- requestFeedbackMutation feedbackFrom =
+--     Reader.Reader
+--         (\( env, apiKey ) ->
+--             GQLClient.sendMutation (graphqlUrl env) <|
+--                 GQLBuilder.request feedbackFrom feedbackMutation
+--         )
```
vipulnsward: Lets get rid of this

Issue #206, opened 2018-07-17
Library page not working


vipulnsward:
https://monosnap.com/file/LKgqAiIN3cOHDdv4xG6iTSjKWJXWcl

@tejaswinichile _a

PR #208, opened 2018-07-17
library page issue resolved

tejaswinichile:
1. fixed nested categories query
2. inserted article data in categories

Issue #206


vipulnsward:
@breezyboa _a

Issue #214, opened 2018-07-18
Bug: Clicking on "New Article", "New Url", "New Category" buttons reloads the page


breezyboa:
Defeats the purpose of SPA

Call `NavigateTo Page` Message onclick instead of using href

PR #217, opened 2018-07-18
navigate to new page urls

tejaswinichile:
> Navigate pages instead of reloading
> created a common component to render page URLs

Issue #214


app/javascript/Admin/Page/Common/Routing.elm
```diff
@@ -0,0 +1,24 @@
+module Page.Common.Routing exposing (..)
```
breezyboa: No need to create this file use the existing Route module
use the existing `routeToString` function get the url

tejaswinichile:
@breezyboa _a Please have a look.

Issue #178, opened 2018-07-02
Show/Edit page for an article


neerajdotname:
https://www.figma.com/file/BM6gbtunF52KloxgsFilOT/Show-Page-of-an-Article-AceHelp

It's still WIP but if someone is short on issues then one can get an idea of what it is doing.

vipulnsward:
@sudeeptarlekar _A

vipulnsward:
@breezyboa _a

breezyboa:
@neerajdotname could you clarify the behavior of "Draft" and "Published" are these just indicators?

https://monosnap.com/file/wdkOXAbSGyDAIPjJbFFx9sbQO6uApz

neerajdotname:
Draft means article is wip. It should not be visible to anyone using the widget.

Published articles should be visible to everyone.

PR #237, opened 2018-07-24
Show an Article

breezyboa:
Part of #178


breezyboa:
@sudeeptarlekar _a can you review this please

PR #255, opened 2018-07-26
shorter method name

neerajdotname:


neerajdotname:
@breezyboa What needs to be done here to fix circleci issue?

Issue #249, opened 2018-07-25
Handle show articles route in rails


breezyboa:
`/admin/articles/4ba7ddab-defc-4156-beb9-bb4a38618b81`
Should open Article with id `4ba7ddab-defc-4156-beb9-bb4a38618b81`

Elm side routes are taken care of in #237

tejaswinichile:
@breezyboa Right now, its opening `edit` page, instead of that we should show page?
I am trying to understand an issue here.

PR #275, opened 2018-07-27
render article show page after reload

tejaswinichile:
Issue #249


breezyboa:
Looks straight forward. @sudeeptarlekar _a can you still review the Ruby code once

PR #277, opened 2018-07-27
Ability to add and remove category from article

### LINKED ISSUE #272: Build endpoint to add category and to remove category from an article
```
addCategoryToArticle  : accepts category_id and article_id
removeCategoryFromArticle : accepts category_id and article_id
```

neerajdotname:
fixes #272 

@sudeeptarlekar _a Can you implement remove category from article ?


sudeeptarlekar:
@neerajdotname Sure will continue in this PR

app/graphql/mutations/category_article_mutations.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+class Mutations::CategoryArticleMutations
+  AddCategoryToArticle = GraphQL::Relay::Mutation.define do
+    name "AddCategoryToArticle"
```
sudeeptarlekar: shouldn't this be `AddArticleToCategory` because
  article belongs to category and category has many articles.

@neerajdotname let me know your thoughts on this...

neerajdotname: I named it because in the edit page for article one can add more categories and remove categories.

So when you are in the edit article screen then you are "adding a category" or "removing a category".

sudeeptarlekar: I see, Thanks

sudeeptarlekar:
@neerajdotname _a added mutations, Please review

app/graphql/mutations/category_article_mutations.rb
```diff
@@ -0,0 +1,77 @@
...
+    return_field :errors, types[Types::ErrorType]
+
+    resolve -> (object, inputs, context) {
+      category = Category.for_organization(context[:organization])
+                         .where(id: inputs[:category_id]).first
+
+      article = Article.for_organization(context[:organization])
+                         .where(id: inputs[:article_id]).first
+
+      if category.blank?
```
neerajdotname: I should be able to remove `category_id` from an article even if that category does not exist. In fact that could be a good reason why I might want to remove a zombie `category_id`.

However don't worry about it. I'll fix it.

neerajdotname:
I just realized that an article can belong to only one category and that's not ideal.

I looked at help docs for one of our clients and there are many articles which are present in two or more categories.

We will take up this work of having an article belong to many categories in another issue.

PR #289, opened 2018-07-28
Long lines make it harder to read code in laptop

neerajdotname:
This commit has only indentation changes.
No business logic is changed.


vipulnsward:
Hmm, I thought elm-format already did this for us?

neerajdotname:
It does. That's how I did this change. Just break the line anywhere and save the file and it will be formatted.

However it behaves differently than prettier. Prettier also takes into account width of the screen. I'm not sure what's the default screen width prettier has.

In Elm if we have a really really long line (high width) then elm will not break that. If we break the line then elm will format it.

PR #296, opened 2018-07-30
252 redirect user to organization create after signup

abdul-shajin:
Serverside route changes
2 New Controllers, Registration and Organizations
Elm routes, pages, GraphQl mutations
Form, Form validation
Controller tests

### DIFF
--- a/app/javascript/Admin/Admin/Request/Organization.elm
+++ b/app/javascript/Admin/Admin/Request/Organization.elm
@@ -44,7 +44,7 @@ requestOrganization env apiKey organizationId =
             , withCredentials = False
             }
 
-requestCreateOrganization : OrganizationData -> Reader NodeEnv (Task GQLClient.Error OrganizationData)
+requestCreateOrganization : OrganizationData -> Reader NodeEnv (Task GQLClient.Error Organization)
 requestCreateOrganization orgInputs =
     Reader.Reader
         (\env ->
--- a/app/javascript/Admin/Admin/Request/Organization.elm
+++ b/app/javascript/Admin/Admin/Request/Organization.elm
@@ -52,7 +52,7 @@ requestCreateOrganization orgInputs =
                 (GQLBuilder.request
                     { name = orgInputs.name
                     , email = orgInputs.email
-                    , user_id = orgInputs.user_id
+                    , userId = orgInputs.userId
                     }
                     createOrganizationMutation
                 )
--- a/app/javascript/Admin/Admin/Request/Organization.elm
+++ b/app/javascript/Admin/Admin/Request/Organization.elm
@@ -1,11 +1,14 @@
-module Request.OrganizationRequest exposing (..)
+module Admin.Request.Organization exposing (..)
 
 import Http
 import Json.Decode as JD exposing (field)
 import Json.Encode as JE
 import Admin.Request.Helper exposing (..)
 import Admin.Data.Organization as AD exposing (..)
-
+import Reader exposing (Reader)
+import Task exposing (Task)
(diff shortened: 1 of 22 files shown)


app/javascript/Admin/Admin/Request/Organization.elm
```diff
@@ -40,3 +43,17 @@ requestOrganization env apiKey organizationId =
             , timeout = Nothing
             , withCredentials = False
             }
+
+requestCreateOrganization : OrganizationData -> Reader NodeEnv (Task GQLClient.Error Organization)
+requestCreateOrganization orgInputs =
+    Reader.Reader
+        (\env ->
+            GQLClient.sendMutation (graphqlUrl env) <|
```
sudeeptarlekar: Ok, here is a catch, this mutation might work in development mode but it will fail on staging, as you are not passing organization API key. Use `customSendMutation` instead of `sendMutation`.

neerajdotname: @sudeeptarlekar Why is that it will be work in development and while it won't work in staging. We should fix our tool chain so that something that will break in staging will also break in development. Can you create another issue and do a fix there.

app/javascript/Admin/Page/Organization/Create.elm
```diff
@@ -0,0 +1,167 @@
...
+import Navigation
+import Route
+
+-- MODEL
+
+type alias Model =
+    { error: Maybe String
+    , name : Field String String
+    , email: Field String String
+    , user_id: String
```
sudeeptarlekar: Please use camel case pattern everywhere

app/javascript/Admin/Admin/Request/Organization.elm
abdul-shajin: but the mutation I'm sending here is to create the organization right after the Signup. The user will not be any part of Organization.

sudeeptarlekar: Problem is in `load_organization.rb` concern where we are loading first organization by default if in development mode and this is suppressing the error on local.

abdul-shajin:
Have fixed the issue in another PR, Would rebase with the changes once it got merged and close the PR

abdul-shajin:
@sudeeptarlekar _a

Rebased with the load_organization.rb changes, api-key in header will not be an issue now.

app/javascript/Admin/Page/Organization/Create.elm
sudeeptarlekar: @abdul-shajin _a please fix this

abdul-shajin: @sudeeptarlekar _a done that

app/javascript/Admin/Page/Organization/Create.elm
```diff
@@ -0,0 +1,167 @@
...
+            in
+                if isAllValid fields then
+                    save model nodeEnv
+                else
+                    ( {model | error = Just "Please check your inputs" }, Cmd.none )
+        SaveOrgResponse (Ok id) ->
+            ({ model
+            | name = Field.update model.name ""
+            , error = Nothing
+            }, Navigation.modifyUrl (Route.routeToString (Route.ArticleList id.id)))
```
tejaswinichile: We are using organization's API key here. not ID. Can you check that because of it's breaking after saving https://monosnap.com/file/uNRTbChvWXbV3O1S6CQtfU4vKxGwbj

tejaswinichile:
@abdul-shajin _a

Issue #312, opened 2018-07-31
Add user to the organization


neerajdotname:
GraphQL endpoint to add a new user to the organization.

accepts : email (required), name(optional). 

If a user with given email already exists then add that user to the org.

If a user with that email does not exist then create a user with that email and then add that user to the org.

abdul-shajin:
FYI,
When we have to create a user, password field have to be filled and for now, I have given a token generated by Devise itself. 
When we handle the complete Onboarding process, we may have to deal this differently

tejaswinichile:
@abdul-shajin Then we should send password reset mail to the user, right?
cc : @neerajdotname

neerajdotname:
See what AceInvoice does. Do the same.

tejaswinichile:
In AceInvoice they do send one welcome email after adding of the user. @abdul-shajin  Please check this and make necessary changes.

PR #325, opened 2018-07-31
GraphQL API:Adding user to organization

abdul-shajin:
GraphQL endpoint to add a new user to the organization.
accepts : email (required), name(optional).
If a user with given email already exists then add that user to the org.
If a user with that email does not exist then create a user with that email and then add that user to the org.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -19,10 +19,27 @@ class User < ApplicationRecord
     "#{first_name} #{last_name}".squish
   end
 
+  def name=(name)
+    user = self
+    user.first_name, user.last_name = name.split(/\s+/, 2)
+  end
+
+  def extract_firstname_from_email
+    email.split('@').first
+  end
+
   def add_organization(args)
     user = self
     user.organization = Organization.new(email: args["email"], name: args["name"])
     user.save
     user.organization
   end
+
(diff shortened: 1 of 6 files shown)


app/models/user.rb
```diff
@@ -19,10 +19,27 @@ def name
     "#{first_name} #{last_name}".squish
   end
 
+  def name=(name)
+    user = self
+    user.first_name, user.last_name = name.split(/\s+/, 2)
+  end
+
+  def extract_firstname_from_email
```
tejaswinichile: Why do we need to extract this?

abdul-shajin: because first_name in user model is mandatory and in graphql, the name field is optional

tejaswinichile: Then I guess we should have same validation at graphQL side and model.
@neerajdotname Need your thoughts on this. Do you know why we are keeping these different validation at different level?

neerajdotname: We need to fix the model. First name should not be required. If user does not have name then anywhere we show user name we should show `Anonymous`.

tejaswinichile:
LGTM 👍

tejaswinichile:
#312

PR #333, opened 2018-08-01
Better organization vadlidation

abdul-shajin:
GraphQL should raise exception if someone access context[:organization] without sending proper header


app/controllers/graphql_controller.rb
```diff
@@ -61,4 +61,19 @@ def show_error_in_logs(e)
 
       raise e if Rails.env.test?
     end
+
+    def find_org
+      api_key = request.headers["api-key"] || params["organization_api_key"]
+      Organization.find_by(api_key: api_key)
+    end
+
+    def hash_raise_if_no_org
```
neerajdotname: @abdul-shajin _a Please change the method to be more explicit. Something like

`custom_hash_that_raises_exception_when_no_org_is_found`.

app/controllers/graphql_controller.rb
```diff
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 class GraphqlController < ApplicationController
-  include LoadOrganization
```
neerajdotname: Not sure if this module is being used by anyone else. Most likely not. In that case this PR should also delete the module.

PR #334, opened 2018-08-01
Remove user from oragnization

abdul-shajin:
Remove user from **oragnization**
Created new client for GraphQL test

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -38,8 +38,6 @@ class User < ApplicationRecord
   end
 
   def deallocate_from_organization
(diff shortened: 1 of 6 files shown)


app/models/user.rb
```diff
@@ -25,4 +25,10 @@ def add_organization(args)
     user.save
     user.organization
   end
+
+  def deallocate_from_organization
+    user = self
```
tejaswinichile: Do we need to assign it explicitly? You can use `update_attributes` here.

Issue #318, opened 2018-07-31
Accept full user agent and process it in background


neerajdotname:
Ref https://github.com/bigbinary/acehelp/issues/317

Store "user agent" in a column. Using background job process this "user agent" and populate browser info etc in json format in a column so that this info can be used by agents to better understand what browser/machine user was using.

abdul-shajin:
Just giving my understandings,
We need to add a column `user_agent` in tickets table and need to fill the data every time user creates a new ticket. And parse that using a background process and save it in a different table with details such as, OS, browser, phone/ipad and so. And this record will be associated with a ticket.

neerajdotname:
> save it in a different table

No. Save it in a different column.

PR #351, opened 2018-08-03
Parse user agent and save in device_info

abdul-shajin:
@tejaswinichile _a


tejaswinichile:
@abdul-shajin _a please check and resolve conflicts.

abdul-shajin:
@tejaswinichile  _a Resolved conflicts

app/controllers/graphql_controller.rb
```diff
@@ -27,8 +27,12 @@ def operation_name
     end
 
     def context
-      {}
-      { organization: @organization } if @organization.present?
+      context = {}
+      context[:organization] = @organization if @organization.present?
+      if request_is_mutation_for?("addTicket")
+        context[:user_agent] = request.user_agent
```
tejaswinichile: Do we need to handle in context?
Can we pass it in specific ticket mutation?

tejaswinichile:
@abdul-shajin _a

tejaswinichile:
#318

app/controllers/graphql_controller.rb
abdul-shajin: How do we do that?

We need to pass a variable from GraphQL#execute to mutation file. and request object doesn't have access in mutation files.

Issue #270, opened 2018-07-27
Ability to add a new category


neerajdotname:
@Shrikant9 _a Please work on this after you are done with https://github.com/bigbinary/acehelp/issues/268

PR #354, opened 2018-08-07
Adds graphql integration for create category functionality.

Shrikant9:
Issue [#270](https://github.com/bigbinary/acehelp/issues/270) Adds create category integration.

### DIFF
--- a/app/javascript/Admin/Page/Category/Create.elm
+++ b/app/javascript/Admin/Page/Category/Create.elm
@@ -9,9 +9,9 @@ import Helpers exposing (..)
 import Html exposing (..)
 import Html.Attributes exposing (..)
 import Html.Events exposing (..)
-import Navigation exposing (modifyUrl)
 import Reader exposing (Reader)
 import Request.Helpers exposing (..)
+import Route
 import Task exposing (Task)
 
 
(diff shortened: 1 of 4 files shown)


Shrikant9:
@breezyboa _a

app/javascript/Admin/Page/Category/Create.elm
```diff
@@ -82,24 +86,11 @@ update msg model nodeEnv organizationKey =
                 , name = Field.update model.name ""
                 , error = Nothing
               }
-            , Cmd.none
+            , modifyUrl ("/organizations/" ++ organizationKey ++ "/categories")
```
breezyboa: This behavior is a little strange. The user will not understand if a category has been created since we are not showing any message of either successful category creation or an error. Maybe a modal view in the category listing page is better in this case to create a category.

This goes for others as well i.e Url and maybe even an Article but we can look at that separately.

I think a better way to do this might be to not have a separate category creation page at all and rather just show a modal view. It would be better that way since creating a category just requires a category name

We can take care of this in a separate PR

neerajdotname:
I’m not following the comment’s conclusion.

Is it asking to do things differently or not. 

Either way for now let’s follow the same pattern we have been using for creating other resources.

app/javascript/Admin/Page/Category/Create.elm
breezyboa: can you use `Route.modifyUrl` instead. do not use strings directly here
Look at line 90 in `Page.Url.Create` module for reference

breezyboa:
@neerajdotname yes I am thinking we do it in a different way. I can create separate issues for this. But for now this is okay since its how we have been doing it in other resources

@Shrikant9 _a please address the second comment

Shrikant9:
@breezyboa _a 
Updated the redirection with `Route.modifyUrl`. Please check.

Issue #329, opened 2018-08-01
Auto close "Resolved" tickets in 4 days.


neerajdotname:
https://github.com/bigbinary/aceinvoice-web/blob/428f70f39759741a8d5f447779b0e88adf484436/Gemfile#L60

Ref https://github.com/bigbinary/acehelp/wiki/Triggers-for-Tickets

abdul-shajin:
Closing as there's PR open

PR #361, opened 2018-08-07
Auto close resolved tickets for 4 days

abdul-shajin:
Issue #329

### DIFF
--- a/app/models/ticket.rb
+++ b/app/models/ticket.rb
@@ -31,4 +31,5 @@ class Ticket < ApplicationRecord
       end
     end
     handle_asynchronously :parse_user_agent
+
 end
--- a/app/models/ticket.rb
+++ b/app/models/ticket.rb
@@ -8,20 +8,19 @@ class Ticket < ApplicationRecord
   has_many :comments, dependent: :destroy
 
   scope :for_organization, ->(org) { where(organization: org) }
-  scope :all_closed, -> { where(status:  STATUSES[:closed]) }
-  scope :all_resolved, -> { where(status:  STATUSES[:resolved]) }
-  scope :all_resolved_before_n_days, -> (day_count) { all_resolved.where('resolved_at < ?', day_count.days.ago) }
+  scope :all_resolved_before_n_days, -> (day_count) { resolved.where('resolved_at < ?', day_count.days.ago) }
 
   after_save :parse_user_agent, if: :saved_change_to_user_agent?
   after_save :mark_status_updates, if: :saved_change_to_status?
 
-  STATUSES = {
-    open: "open",
-    pending_on_customer: 'pending_on_customer',
-    resolved: 'resolved',
-    closed: 'closed'
+  enum status: {
+      open: "open",
+      pending_on_customer: 'pending_on_customer',
+      resolved: 'resolved',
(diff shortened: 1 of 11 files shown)


app/models/ticket.rb
```diff
@@ -8,8 +8,20 @@ class Ticket < ApplicationRecord
...
 
   scope :for_organization, ->(org) { where(organization: org) }
+  scope :all_closed, -> { where(status:  STATUSES[:closed]) }
+  scope :all_resolved, -> { where(status:  STATUSES[:resolved]) }
+  scope :all_resolved_before_n_days, -> (day_count) { all_resolved.where('resolved_at < ?', day_count.days.ago) }
 
   after_save :parse_user_agent, if: :saved_change_to_user_agent?
+  after_save :mark_status_updates, if: :saved_change_to_status?
+
+  STATUSES = {
```
tejaswinichile: You can declare it with an enum. So that you will be able to use it as `close!` to close the ticket.

app/models/ticket.rb
```diff
@@ -31,4 +52,17 @@ def parse_user_agent
       end
     end
     handle_asynchronously :parse_user_agent
+
+    def mark_status_updates
+      if saved_change_to_status?
+        if status == STATUSES[:resolved]
```
tejaswinichile: With `enum` declaration you can use `resolved?`

tejaswinichile:
We will need to set up delayed job on Heroku after this PR is merged to get it to work.
I checked, seems the worker process is in OFF state on Heroku.

tejaswinichile:
@abdul-shajin _a

app/models/ticket.rb
abdul-shajin: Enums are okay if we just think ActiveRecord perceptive, I had to change entire enum attrs to string in past when we had started to add complex queries and when started using third party apps to get the reports  etc . Since PostgreSQL or any db doesn't know what does 1,2,3 means.

app/models/ticket.rb
abdul-shajin: Possible, but see my previous comment

app/models/ticket.rb
tejaswinichile: @abdul-shajin You can add the different keys for enum, instead of 1, 2, 3.

abdul-shajin: great, didnt know that.. will make the change

abdul-shajin:
Done review changes
@tejaswinichile _a

app/models/ticket.rb
```diff
@@ -21,6 +32,12 @@ def add_note(note_txt)
     update(note: note_txt)
   end
 
+  def update_status(status_key)
+    return false if status == status_key.to_s
```
tejaswinichile: where are we using this? And any significance for returning `false`?

tejaswinichile:
@abdul-shajin one small comment please have a look. _a

app/models/ticket.rb
abdul-shajin: No, Just a way of programming to make sure both 'open' and :open to work..
and return (nil/false) to make sure the execution of the function to be stopped there..
what do yo propose?
@tejaswinichile _a

Issue #328, opened 2018-08-01
Ticket edit view


neerajdotname:
![img_9337](https://user-images.githubusercontent.com/6399/43501593-e0a2340a-9573-11e8-8fc8-08b6f3621dad.JPG)

neerajdotname:
A ticket can have multiple notes too.

tejaswinichile:
@neerajdotname why multiple notes? Only the agent to whom the ticker is assigned can post the note, right? And other agents can see that note.
I am trying to understand requirement here.

tejaswinichile:
Q: 
1.  How many ticket statuses are there?
2. Assigned to dropdown >> Will it have the current username?
3. Close ticket >> Does that mean ticket resolved?
4. Delete ticket >> Any state should current ticket contain before deleting like ticket should be always in the resolved state before delete.

@neerajdotname Have few questions.

neerajdotname:
Look at ticket model or something like that to see all possible state. Roughly it is open, pending to customer, resolved and closed.

Resolved means agent is confident that agent’s answer resolves the issue. If customer does not reply then ticket will be auto closed. It is handled in another issue.

We do not have concept of username. Drop down will have the names of agents.

Delete ticket is not a status of ticket. This is to mark a ticket which should not even appear in any report because it is either spammy or totally useless issue. A ticket should have deleted_at column to mark if a ticket is deleted it not. Any ticket in any state can be deleted.

tejaswinichile:
- [ ] Render ticket edit page with minimal options

 - [ ] Update Ticket status

- [ ] Close Ticket

- [ ] Deleted Ticket

- [ ] Toggle between ticket note and ticket comment

- [ ] Update ticket with ticket note

- [ ] Update ticket with ticket comment

tejaswinichile:
@neerajdotname Need confirmation here, as mentioned here https://github.com/bigbinary/acehelp/issues/327 << any agent can add notes to the ticket, right?
Or do we need to add the condition to check if the ticket is assigned to that agent or not?

neerajdotname:
Any agent can add notes.

neerajdotname:
Auto assign should be applied only to the ticket to which no agent is assigned. Once a ticket is assigned to an agent then there is not auto assignment. Other agents can comment but it won't assign ticket to them.

Once a ticket has been assigned to an agent then the only way to assign it to some other agent is by someone taking an action. In the UI someone clicks on change agent and selects from the list of presented agents.

PR #364, opened 2018-08-08
Change ticket status

tejaswinichile:
Issue #328

### DIFF
--- a/app/graphql/mutations/change_ticket_status_mutation.rb
+++ b/app/graphql/mutations/change_ticket_status_mutation.rb
@@ -4,15 +4,15 @@ class Mutations::ChangeTicketStatusMutation
   Perform = GraphQL::Relay::Mutation.define do
     name "ChangeTicketStatus"
 
-    input_field :status, !Types::TicketStatusEnumType
-    input_field :ticket_id, !types.String
+    input_field :status, !types.String
+    input_field :id, !types.String
 
     return_field :ticket, Types::TicketType
     return_field :errors, types[Types::ErrorType]
 
     resolve ->(object, inputs, context) {
 
-      ticket = Ticket.find_by(id: inputs[:ticket_id])
+      ticket = Ticket.find_by(id: inputs[:id])
 
       if ticket
         updated_ticket = ticket if ticket.update(status: inputs[:status])
--- a/app/javascript/Admin/Admin/Data/Ticket.elm
+++ b/app/javascript/Admin/Admin/Data/Ticket.elm
@@ -73,21 +73,7 @@ requestTicketByIdQuery =
             (GQLBuilder.extract
                 (GQLBuilder.field "ticket"
                     [ ( "id", Arg.variable idVar ) ]
-                    (GQLBuilder.object TicketEditData
-                        |> GQLBuilder.with (GQLBuilder.field "id" [] GQLBuilder.string)
-                        |> GQLBuilder.with (GQLBuilder.field "name" [] GQLBuilder.string)
-                        |> GQLBuilder.with (GQLBuilder.field "email" [] GQLBuilder.string)
-                        |> GQLBuilder.with (GQLBuilder.field "message" [] GQLBuilder.string)
-                        |> GQLBuilder.with (GQLBuilder.field "note" [] GQLBuilder.string)
-                        |> GQLBuilder.with (GQLBuilder.field "status" [] GQLBuilder.string)
-                        |> GQLBuilder.with
-                            (GQLBuilder.field "statuses"
-                                []
-                                (GQLBuilder.list
-                                    ticketStatusObject
-                                )
-                            )
-                    )
+                    (ticketObject)
                 )
             )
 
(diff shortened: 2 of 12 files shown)


tejaswinichile:
@breezyboa _a

app/javascript/Admin/Admin/Data/Ticket.elm
```diff
@@ -13,6 +13,38 @@ type alias Ticket =
...
+    }
+
+
+type alias TicketId =
+    String
+
+
+type alias TicketIdInput =
+    { id : String
+    }
```
breezyboa:
```
type alias TicketIdInput =
    { id : TicketId
    }
```

you could even define this in place if it is never going to be resused

eg:
```
updateTicket : Reader ( NodeEnv, ApiKey,{ id : TicketId} ) (Task GQLClient.Error TicketEditData)
```

app/javascript/Admin/Admin/Data/Ticket.elm
```diff
@@ -13,6 +13,38 @@ type alias Ticket =
...
+
+type alias TicketEditData =
+    { id : String
+    , name : String
+    , email : String
+    , message : String
+    , note : String
+    , status : String
+    , statuses : List TicketStatus
+    }
```
breezyboa: Why does this type have a `status` and a `statuses` field?
Why not include `status` in the `Ticket` type?

app/javascript/Admin/Page/Ticket/Edit.elm
```diff
@@ -0,0 +1,233 @@
...
+
+
+type alias Model =
+    { error : Maybe String
+    , success : Maybe String
+    , note : String
+    , message : String
+    , ticketId : TicketId
+    , status : String
+    , statuses : List TicketStatus
```
breezyboa: Again why do we have status and statuses?
If the idea is to have a list of statuses and one particular status needs to be selected. Then it would be better to represent this using a Zipper List. This will avoid impossible states.  Check Client/Views/Tabs.elm for implementation

app/javascript/Admin/Admin/Data/Ticket.elm
tejaswinichile: We will be reusing that, in the same ticket only. We are gonna have multiple functions associated to this. Like `deleteTciket` `closeTicket` `addNote` `addComment`.

app/javascript/Admin/Page/Ticket/Edit.elm
tejaswinichile: @breezyboa 
I saw zipper implementation, the thing which we are doing in `TabSelected tab` to set the current tab `onClick`. 

But here, I need to fetch that selected `status` which is already saved from the server that's what the purpose of having `status` and `statuses` just provides various status options.

breezyboa: The point is that the model allows for a state where `status` can have an entry that may not be present in `statuses` and what does it mean when there exists a value in `status` but `statuses` is empty?

If not with zipper we can also wrap a `status` with an `Option` type from `Admin.Data.Common`
Check Page.Article.Edit or Page.Article.Create for implementation details. In this case statuses would be like:

```
statuses : List (Option TicketStatus)
```

tejaswinichile: Makes sense, let me check those 👍

tejaswinichile: @breezyboa Still I will need to have status and statuses fetch from the server, as we did for `categories` and selected `categoryIds` in `article/edit` page.

tejaswinichile: Please check I have handled with a single selection option.

tejaswinichile:
@breezyboa _a

app/javascript/Admin/Admin/Data/Ticket.elm
```diff
@@ -20,12 +44,73 @@ requestTicketQuery =
...
+ticketObject =
+    GQLBuilder.object Ticket
+        |> GQLBuilder.with (GQLBuilder.field "id" [] GQLBuilder.string)
+        |> GQLBuilder.with (GQLBuilder.field "name" [] GQLBuilder.string)
+        |> GQLBuilder.with (GQLBuilder.field "email" [] GQLBuilder.string)
+        |> GQLBuilder.with (GQLBuilder.field "message" [] GQLBuilder.string)
+        |> GQLBuilder.with (GQLBuilder.field "note" [] GQLBuilder.string)
+        |> GQLBuilder.with (GQLBuilder.field "status" [] GQLBuilder.string)
+        |> GQLBuilder.with
+            (GQLBuilder.field "statuses"
```
breezyboa: Again I have the same question here. Why does a ticket have list of `statuses` when it can only have one status.

If this is about fetching all possible `statuses` from the server should it not be a separate query?

app/javascript/Admin/Page/Ticket/Edit.elm
breezyboa: ? the code is still the same

breezyboa:
@abdul-shajin _a can you review rails code

app/javascript/Admin/Page/Ticket/Edit.elm
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+
+save : Model -> NodeEnv -> ApiKey -> ( Model, Cmd Msg )
+save model nodeEnv organizationKey =
+    --let
+    --    cmd =
+    --        Task.attempt UpdateTicketResponse (Reader.run (updateUrl) ( nodeEnv, organizationKey, urlInputs {  } ))
+    --in
+    ( model, Cmd.none )
```
breezyboa: Looks like this function is not in use. Can you remove it

tejaswinichile: I wanted to use in the next steps of this issue. Where user can update notes and comments.
Removing for now.

app/graphql/mutations/change_ticket_status_mutation.rb
```diff
@@ -4,15 +4,15 @@ class Mutations::ChangeTicketStatusMutation
   Perform = GraphQL::Relay::Mutation.define do
     name "ChangeTicketStatus"
 
-    input_field :status, !Types::TicketStatusEnumType
-    input_field :ticket_id, !types.String
+    input_field :status, !types.String
```
abdul-shajin: we added enum type to make sure status from client should be a valid one..
Dont we need to handle that now?

@tejaswinichile _a

abdul-shajin: Okay, by changing to enum type would solve that issue I think..

abdul-shajin:
Good to go. :100:

app/graphql/mutations/change_ticket_status_mutation.rb
tejaswinichile: Yeah...Enum at model level...it's handling this case. If we pass anything other than defined in Enum throwing the proper error too.

breezyboa:
@tejaswinichile _a please resolve conflicts and merge

neerajdotname:
I noticed a few comments had double dots. Please note double dots might be acceptable in chat but it’s not grammatically correct sentence.

Issue #368, opened 2018-08-08
Single route to maintain SPA


tejaswinichile:
1. have only one route to render the app 
2. get rid of all controllers and actions which we have added for reloading the page

PR #369, opened 2018-08-09
Handle single route for SPA

tejaswinichile:
Issue #368


tejaswinichile:
@abdul-shajin _a Please check.
@vipulnsward Need your thoughts too on same.

abdul-shajin:
LGTM :100:

tejaswinichile:
@vipulnsward _a Please have a look.

app/controllers/application_controller.rb
```diff
@@ -23,7 +23,7 @@ def configure_permitted_parameters
 
     def after_sign_in_path_for(resource)
       if resource.organizations.exists?
-        organization_articles_path(resource.organizations.first.api_key)
+        "/organizations/#{resource.organizations.first.api_key}/articles"
```
vipulnsward: Please use path helper, no handcrafting of urls

config/routes.rb
```diff
@@ -16,12 +16,9 @@
...
 
   resources :organizations, only: [:show], param: :api_key do
-    resources :articles
-    resources :urls
-    resources :categories
-    resources :tickets, only: [:index]
-    resources :feedbacks, only: [:index, :show]
-    resources :team, only: [:index]
+    get "*path", to: "admin/dashboard#index", constraints: -> (request) do
+      !request.xhr? && request.format.html?
```
vipulnsward: What happens if it indeed is an xhr? and html

vipulnsward:
@breezyboa _a please take this up to see how it fits with FE routing

config/routes.rb
tejaswinichile: An XHR request should always be handled by graphQL, not by this route.

breezyboa:
This will handle all routes in `http://localhost:3000/organizations/*`via the frontend

To make this completely work we need all 404s also to be directed to a default page. it could be a root path like `http://localhost:3000/`. Let's take care of that separately

Merging this PR

Issue #353, opened 2018-08-06
updateArticle mutation throws a 500 error


breezyboa:
updateArticle mutation throws a 500 error

- Article gets updated even though error is thrown
- the mutation does not return the updated article back
- the mutation does not allow urls to be assigned/updated

https://monosnap.com/file/55I95Jhr9X03AGTFvlTfO2NfG1oQYE

Graphql mutation:
```
mutation ($id: String!, $title: String!, $desc: String!, $category_id: String = "") {
  updateArticle(input: {id: $id, title: $title, desc: $desc, category_id: $category_id}) {
    article {
      id
      title
      desc
      categories {
        id
        name
      }
      urls {
        id
        url
      }
    }
  }
}
```
variables:
```
{id: "9700297b-d648-4956-8798-4e1934b45fff", title: "How do I put JavaScript code in my website?", desc: "<div><!--block-->Article details is coming soon</div>"}
```

tejaswinichile:
@breezyboa We don't have update button for article edit page, it's getting updated on change event of fields. Is that correct implementation?

breezyboa:
https://www.figma.com/file/BM6gbtunF52KloxgsFilOT/Show-Page-of-an-Article-%7C-AceHelp?node-id=0%3A1

You are right the implementation needs to be corrected. The auto save feature is only when user selects a category or url. There should be a save and cancel button for the article contents

I will get it corrected in a separate issue

tejaswinichile:
@breezyboa One more thing I have observed, we are not able to select multiple options at a time. If select first option other option gets unselected.
Whenever we will fix the issue, need to fix the issue for updating categories and URLs associated to the article from the server side. As now user not able to unselect the option and cannot remove associated category or URL from an article.
Fixed all issues mentioned in this ticket though.

breezyboa:
@tejaswinichile multiple options are selected by holding down the control/command key and then selecting multiple items. that's the default behavior of select/option

tejaswinichile:
@breezyboa I think we need to change passing category_id behavior. We need to send all selected category_ids to GQL. Else we will need to pass unselect option, to get the info that this category has been unselected.

We will handle this in the separate issue.

neerajdotname:
> We need to send all selected category_ids to GQL.

I think this is ok. In this case it will be the responsibility of the server to ensure that the category ids not passed in the list are removed.

PR #384, opened 2018-08-10
Fix update article

tejaswinichile:
Issue #353


tejaswinichile:
@breezyboa  _a Please have a look.

tejaswinichile:
@breezyboa I am going to handle that in the new issue.
As we need to change it from rails side and from Elm side too. We can have this PR merged.

tejaswinichile:
@breezyboa _a #390 Here is the issue to handle multiple category_ids, you can merge this PR if everything seems okay.

Issue #332, opened 2018-08-01
GraphQL query : show all triggers


neerajdotname:

tejaswinichile:
Will we be creating these triggers from the background(with setup task)? Or we are giving some kind of dashboard for admin to create triggers.

PR #391, opened 2018-08-13
332-list-triggers: list all triggers

abdul-shajin:
Added model, migrations
Added GraphQL Query
Added GraphQL Tests
Added Model tests

Issue #332

### DIFF
--- a/app/models/trigger.rb
+++ b/app/models/trigger.rb
@@ -1,0 +1,7 @@
+class Trigger < ApplicationRecord
+
+  scope :all_active, -> { where(active: true) }
+  scope :all_inactive, -> { where(active: false) }
+
+
+end
--- a/app/models/trigger.rb
+++ b/app/models/trigger.rb
@@ -1,7 +1,7 @@
 class Trigger < ApplicationRecord
 
-  scope :all_active, -> { where(active: true) }
-  scope :all_inactive, -> { where(active: false) }
+  scope :active, -> { where(active: true) }
+  scope :inactive, -> { where(active: false) }
 
 
 end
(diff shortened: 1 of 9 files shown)


tejaswinichile:
@abdul-shajin _a test cases are failing, can you have a look.

abdul-shajin:
@tejaswinichile _a some tests in resolvers were failing. Fixed

db/migrate/20180807160031_create_triggers.rb
```diff
@@ -0,0 +1,16 @@
+class CreateTriggers < ActiveRecord::Migration[5.2]
+  def change
+    create_table :triggers, id: :uuid do |t|
+      t.string :slug, null: false
```
tejaswinichile: I think this should be unique. I don't know exact requirement here. Can you confirm this on the issue?

tejaswinichile:
@abdul-shajin  _a

db/migrate/20180807160031_create_triggers.rb
neerajdotname: Why do we need a slug for a trigger?

abdul-shajin:
@tejaswinichile  _a  I added unique constraint and the index

neerajdotname:
@abdul-shajin Please do not forget to end sentence with a period. Also in that sentence you missed out "have".

It should have been like this.

> I have added unique constraint and the index.

app/models/trigger.rb
```diff
@@ -0,0 +1,7 @@
+class Trigger < ApplicationRecord
+
+  scope :all_active, -> { where(active: true) }
```
tejaswinichile: Just `active` will be sufficient, instead of `all_active`. Same for `inactive`.

abdul-shajin:
Thanks Neeraj. Will keep that in mind..

abdul-shajin:
@tejaswinichile _a Done the review suggestions.

Issue #393, opened 2018-08-13
Add functionality to fetch user details and update user in Ruby.


Shrikant9:

tejaswinichile:
We don't need to update team member yet.
If we want to update member's role we need to confirm that first.

tejaswinichile:
@breezyboa Can you confirm the above requirement, If we are updating team member's data or not.

tejaswinichile:
@breezyboa _a Need your confirmation on above requirement If we don't need to update team member, close this ticket.

breezyboa:
If this is about updating a users role. We do not need it right now

tejaswinichile:
Closing this issue.

PR #398, opened 2018-08-14
fetch user details

tejaswinichile:
Issue #393

### DIFF
--- a/test/graphql/resolvers/users_search_test.rb
+++ b/test/graphql/resolvers/users_search_test.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+require "test_helper"
+require "graphql/client_host"
+
+class Resolvers::UsersSearchTest < ActiveSupport::TestCase
+  setup do
+    @user = users :brad
+    @organization = organizations :bigbinary
+    @user.organizations << @organization
+  end
+
+  def find(args)
+    Resolvers::UsersSearch.new.call(nil, args, organization: @organization)
+  end
+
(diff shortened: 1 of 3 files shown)


tejaswinichile:
@sudeeptarlekar _a

neerajdotname:
Let's stop assigning issues to Sudeep going forward.

Any Rails work can be assigned to me for review.

test/graphql/resolvers/users_search_test.rb
```diff
@@ -0,0 +1,20 @@
...
+    @organization = organizations :bigbinary
+    @user.organizations << @organization
+  end
+
+  def find(args)
+    Resolvers::UsersSearch.new.call(nil, args, organization: @organization)
+  end
+
+  test "get_all_users_success" do
+    assert_equal [@user.email], find(Hash.new).pluck(:email)
```
neerajdotname: Any particular reason why we are using `Hash.new` instead of `{}`. I have never seen `Hash.new` being used like this. Usually `Hash.new` is used to give that hash some extra features. For a blank hash using `{}` is sufficient.

tejaswinichile: No not a particular reason. Understood the point 👍

tejaswinichile:
@neerajdotname _a

tejaswinichile:
@abdul-shajin _a Can you check this PR?

abdul-shajin:
@tejaswinichile _a LGTM 💯

PR #401, opened 2018-08-14
add ticket commenter

tejaswinichile:
1. render ticket commenter instead of commentable


tejaswinichile:
Merging it on priority, changed according to the new commenter changes.

Issue #395, opened 2018-08-13
Handle all paths via a single rails action


breezyboa:
Currently, paths that match the pattern:`http://localhost:3000/organizations/:string:/*` are handled and are routed to a rails controller action

We need to direct all requests including Login, Signup, Forgot Password and even 404s to a single rails controller action.
Except for the following:
- Graphql calls
- xhr calls
- requests to assets (files like .js, .css, .json, .map, .ico, font files, Images and docs )

tejaswinichile:
- [ ] Handle 404 with Elm page

- [ ] Render all client side route with single rails action

- [ ] Get rid of all the controllers in the project

- [ ]  Fix all the test cases according to it

- [ ] Handle page not found in client-side Elm pages

tejaswinichile:
@breezyboa Can you help me here to understand which elm-hook we are gonna use for all those routes which doesn't belong inside the organization as we can not use the same action for those routes. Are we creating new main file Or the same `admin-hook`?

breezyboa:
@tejaswinichile We can create a separate entry point. But can you help me understand why this is needed?

tejaswinichile:
@breezyboa Not needed. I just wanted to confirm. I have created the separate action for the routes which doesn't need to load organization API-key before rendering.

PR #402, opened 2018-08-14
All route to single action

tejaswinichile:
Issue #395


app/javascript/Admin/Main.elm
```diff
@@ -1031,9 +1003,6 @@ view model =
         NotFound ->
             Errors.notFound
 
-        Blank ->
```
breezyboa: We need a Blank page to display as the app loads up this cannot be removed

tejaswinichile: Okay 👍

tejaswinichile:
@breezyboa _a

breezyboa:
@abdul-shajin _a can you review rails code

PR #404, opened 2018-08-14
Frontend structure overhaul

### LINKED ISSUE #397: Show a "Loading" indicator when navigating from one page to the other

breezyboa:
Fixes #397


breezyboa:
- Once this is merged there will be no need to pass apiKey and nodeEnv around.
- The type `ReaderCmd` and its helper functions will handle it for us
- Some grpahql calls require both env and apiKey such calls can be wrapped with the `Strict` constructor
- Some grpahql calls require both just the env such calls can be wrapped with the `Open` constructor
- Any sort of navigation should be handled in the main module with the `NavigateTo` method. This is to avoid a page refresh

breezyboa:
@tejaswinichile _a can you review this?

app/javascript/Admin/Admin/Data/ReaderCmd.elm
```diff
@@ -0,0 +1,36 @@
...
+map fn readerCmd =
+    case readerCmd of
+        Strict reader ->
+            Strict <| Reader.map fn reader
+
+        Open reader ->
+            Open <| Reader.map fn reader
+
+
+readerCmdToCmd : NodeEnv -> ApiKey -> (a -> msg) -> List (ReaderCmd a) -> Cmd msg
```
tejaswinichile: SO instead of passing Reader cmd for nodeEnv and ApiKey, we will be using this one?

app/javascript/Admin/Page/Article/Create.elm
```diff
@@ -68,50 +69,51 @@ type Msg
     | UrlSelected (List UrlId)
 
 
-update : Msg -> Model -> NodeEnv -> ApiKey -> ( Model, Cmd Msg )
-update msg model nodeEnv organizationKey =
+update : Msg -> Model -> ( Model, List (ReaderCmd Msg) )
```
tejaswinichile: Why the list of `ReaderCmd`? We have only one `Cmd` returned.

tejaswinichile:
@breezyboa I will check, it's impact on my local machine.

app/javascript/Admin/Admin/Data/ReaderCmd.elm
breezyboa: Yes. The update function in main has a partially applied version of this in which we pass the nodeEnv and apiKey only once. Its all wired up

app/javascript/Admin/Page/Article/Create.elm
breezyboa: It is essential that all pages follow the same format to keep the structure predictable and simple, unless and until there is an exception.
We give back a list so that if we want to represent no Cmds then we simply pass an empty list, if there is one there will be only one ReaderCmd and in certain cases, we need to fetch url list, category list, and article all at once(check the `init` function). In future to there may be cases where we want to execute multiple cmds in an update. This we can pass as list items to be converted to a list of Cmds and then batch run them.

tejaswinichile:
@breezyboa Merging this PR, As I have checked admin side and it's working fine. No impact.
It will be better to get this merge so that we can integrate this in other PRs.

PR #410, opened 2018-08-15
Removed simple_token_authentication

abdul-shajin:
Added devise_token_auth
Added migration file
Overriden registration/sessions controller
Changed Mutation
Added TokenType
Added ApplicationController/User model Concersn
Updated Routes
Updated mutation test


abdul-shajin:
Will raise another one with front-end changes

PR #415, opened 2018-08-17
forgot password page

tejaswinichile:
issue #303


tejaswinichile:
@breezyboa _a Please check.

Issue #406, opened 2018-08-14
Show org name in the nav link at the top


neerajdotname:
https://monosnap.com/file/unLuSvvrf3kauOvfMBUfOvaLSqvaaP

PR #416, opened 2018-08-17
show organisation name in header

tejaswinichile:
Issue #406


tejaswinichile:
@breezyboa _a Please check.

tejaswinichile:
@breezyboa _a Please check.

PR #419, opened 2018-08-20
Update ticket note

tejaswinichile:
Issue #328


tejaswinichile:
@breezyboa _a Please check elm side changes.

breezyboa:
Elm side looks good
@neerajdotname _a can you review rails code

app/graphql/mutations/ticket_mutations.rb
```diff
@@ -75,6 +74,12 @@ class Mutations::TicketMutations
           ticket_id: ticket.id
         ) if inputs[:comment].present?
         new_comment.assign_agent_to_ticket(context[:current_user].id) if new_comment
+
+        Note.add_note(
+          details: inputs[:note],
+          agent_id: context[:current_user].id,
+          ticket_id: ticket.id
+        ) if inputs[:note].present?
```
neerajdotname: @tejaswinichile _a 

1. Please do not put trailing `if` condition. Trailing if is only acceptable in one liner. If the statement is of more than on line the start with `if` condition.

2. What if `Note.add_note` fails. How are doing error handling. I scrolled down to see the code. It is using bang method. In that case the method name should be changed to `Note.add_note!`. Then it would be clear to all parties how error is handled.

tejaswinichile:
@neerajdotname _a Please check.

app/models/note.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class Note < ApplicationRecord
+  belongs_to :agent, class_name: "User"
+  belongs_to :ticket
+
+  def self.add_note!(args)
+    agent = Agent.find_by(id: args[:agent_id])
```
neerajdotname: What happens if `agent` is nil.

Method has a bang. So the expectation is that if something goes wrong then it will raise an exception.

However if the agent is not found then it will silently fail.

Please change `find_by` to `find_by!`.

@tejaswinichile _a

db/migrate/20180813093947_create_notes.rb
```diff
@@ -0,0 +1,15 @@
...
+      t.text :details, null: false
+      t.uuid :agent_id, null: false
+      t.uuid :ticket_id, null: false
+
+      t.timestamps
+    end
+
+    add_foreign_key :notes, :users, column: :agent_id
+    add_foreign_key :notes, :tickets
+
```
neerajdotname: This adds the foreign keys. However the index on the foreign keys is still missing. Please add them.

app/models/note.rb
tejaswinichile: I was checking it manually if an agent is `present`, but yeah this makes more sense.

tejaswinichile:
@neerajdotname _a Please check.

PR #447, opened 2018-08-22
No articles list rendered issue resolved

tejaswinichile:
1. checked the query and elm structure
2. issue was related to sample data
3. added rails host to url
4. updated sample data


abdul-shajin:
Good to go 💯

Issue #450, opened 2018-08-22
Error after user enters organisation name


breezyboa:
Steps to reproduce:
- Sign up for a new account
- Login and enter Organisation name
- Notice the following Graphql error message

https://monosnap.com/file/EAYWA4caRciBxedELcpuxBN0gRv4Kr

tejaswinichile:
Closing this issue as PR is up with fixes.

PR #453, opened 2018-08-22
Fix for new organization creation

tejaswinichile:
Issue #450 
1. set error to nil after successful rendering of article list on api key
2. set organization association dependent destroy

### DIFF
--- a/app/javascript/Admin/Main.elm
+++ b/app/javascript/Admin/Main.elm
@@ -720,6 +720,7 @@ update msg model =
                             in
                                 ( { updatedModel
                                     | organizationKey = org.api_key
+                                    , organizationName = org.name
                                   }
                                 , updatedCmd
                                 )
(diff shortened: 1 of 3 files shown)


tejaswinichile:
@breezyboa _a please check

breezyboa:
@tejaswinichile _a the error is still flashed for less than a second and then goes off. 
Also, note that the organization name is also not displayed on top right.

The question is why does the error happen in the first place when it clearly should not

tejaswinichile:
When we are creating new org we do not have any api-key present, then we send another request to check any available organization associated to the user.

tejaswinichile:
@breezyboa _a can you look into this as https://github.com/bigbinary/acehelp/pull/453/files#diff-2bc8a2b43c1b3cd65e5916c2e7338445R722 not making any change, we need to set orgKey to ReaderCmd,  which is not happening immediately after organization creation and it's sending apiKey nil to article list first request.

breezyboa:
@tejaswinichile okay ill have a look

PR #454, opened 2018-08-23
reindex all the tables before starting the application

tejaswinichile:


tejaswinichile:
@abdul-shajin _a

abdul-shajin:
Good to go 💯

Issue #452, opened 2018-08-22
Category success message stays on forever


breezyboa:
- Create a new category
-  Navigate to different pages
- Notice success message is not cleared

https://monosnap.com/file/kA0YHUXyI1tAGGVpLS3B2OR6DMDvGw

tejaswinichile:
We need to introduce close button, right now I have added removeNotification onClick of message only.
And Yes one thing missed out is to update flashMessage after navigation.

neerajdotname:
Just like AceInvoice success message should auto disappear. That will remove the need for having a close button.

tejaswinichile:
Yeah that can also be done 👍

tejaswinichile:
Closing this issue as PR is up for the same.

PR #455, opened 2018-08-23
Resolved success flash message issue

tejaswinichile:
Issue #452 
1. on navigate initalize flash message to empty
2. added expiry time remove flash message
3. added update current_time function which runs at every second
4. added update expiry_time every 10 seconds

### DIFF
--- a/app/javascript/Admin/Main.elm
+++ b/app/javascript/Admin/Main.elm
@@ -5,6 +5,7 @@ import Html exposing (..)
 import Html.Attributes exposing (..)
 import Html.Events exposing (..)
 import Http
+import Time
 import Navigation exposing (..)
 import Page.Session.Login as Login
 import Page.Session.ForgotPassword as ForgotPassword
@@ -96,7 +97,8 @@ type alias Model =
     , appUrl : String
     , error : Maybe String
     , flashElements : List NotificationElement
-    , nextId : Int
+    , currentTime : Time.Time
+    , expiryTime : Float
     }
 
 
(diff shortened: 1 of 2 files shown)


tejaswinichile:
@breezyboa _a Please check.

app/javascript/Admin/Main.elm
```diff
@@ -305,12 +309,24 @@ update msg model =
         renderFlashMessages message messageType =
             Tuple.mapFirst
                 (\model ->
-                    { model | flashElements = [ NotificationElement message messageType ] }
+                    { model
+                        | flashElements = [ NotificationElement message messageType ]
```
breezyboa: Why have a list when there will be only one element in it always

tejaswinichile: @breezyboa Right now we have only one element for success message, but this type we can use for error messages there we have the list of error.

app/javascript/Admin/Main.elm
```diff
@@ -801,7 +826,10 @@ subscriptions model =
             Sub.map ArticleEditMsg <| ArticleEdit.subscriptions articleEditModel
 
         _ ->
-            Sub.none
+            Sub.batch
+                [ Time.every Time.second UpdateCurrentTime
+                , Time.every (10 * Time.second) (DeleteFlashElement)
+                ]
```
breezyboa: It may not be efficient to be subscribed to Time all the time. We just need it when a notification is displayed. It would be better to use a flag in the model to start off this subscription. (Something like showNotification: Bool)
Or a better alternative would be to use setTimeout. We already have a port for setTimeout why not just use that

tejaswinichile: Okay, I didn't know about that we have a port for setTimeout. Yeah, this will be easy with that then.

tejaswinichile:
@breezyboa _a Please review.

Issue #448, opened 2018-08-22
Once a User Successfully signs up. User should be redirected to login page


breezyboa:

tejaswinichile:
- [x]  redirect to login page after successful sign-up
- [x]  redirect to organization page after login response

remove curent_user check and handle every request with auth token << To be done after #435 gets merged

PR #457, opened 2018-08-23
Adds redirection to login on successful user signup.

Shrikant9:
Issue [#448](https://github.com/bigbinary/acehelp/issues/448)


Shrikant9:
@breezyboa _a

Please note: Since the redirection is done in Elm, the elm equivalent of the login page is opened. 
Although when browsing from home, the RoR page is opened.

PR #460, opened 2018-08-23
Graphql article search

### LINKED ISSUE #441: Searching for articles should happen through GraphQL
Search functionality in the widget is still happening through the old API this needs to be moved to graphql

sudeeptarlekar:
This fixes #441 

@breezyboa _a Please review


breezyboa:
@sudeeptarlekar _a build is failing can you check.
Elm code looks good. But do we need a new query for search can we use the articles query with the `search_string` parameter?

Issue #433, opened 2018-08-21
Mark a Category as online or offline


breezyboa:
Admin users should be able to mark a Category as online or offline.
If a category is marked offline then all articles under the category should be marked offline
If even one article is marked online under a category then the category should be marked online

To be done after #428

tejaswinichile:
- [ ] Category status update mutation -Rails
- [ ] Update Category status - Elm

tejaswinichile:
Closing this issue, we have PR raised for the same.

PR #461, opened 2018-08-24
Update category status elm

tejaswinichile:
Issue #433


tejaswinichile:
@breezyboa _a Please check

Issue #432, opened 2018-08-21
Widget should fetch only articles that are online


breezyboa:
Offline articles should not be visible in widget

breezyboa:
This should be taken up after #428

PR #469, opened 2018-08-27
Widget to fetch online article

tejaswinichile:
Issue #432

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -113,7 +113,7 @@ end
 group :test do
   # Adds support for Capybara system testing and selenium driver
   gem "capybara", ">= 2.15", "< 4.0"
-  gem "selenium-webdriver"
+  # gem "selenium-webdriver"
   # Easy installation and use of chromedriver to run system tests with Chrome
   gem "chromedriver-helper"
   # A friendlier Ruby client for consuming GraphQL-based APIs.
--- a/Gemfile
+++ b/Gemfile
@@ -113,7 +113,6 @@ end
 group :test do
   # Adds support for Capybara system testing and selenium driver
   gem "capybara", ">= 2.15", "< 4.0"
-  # gem "selenium-webdriver"
   # Easy installation and use of chromedriver to run system tests with Chrome
   gem "chromedriver-helper"
   # A friendlier Ruby client for consuming GraphQL-based APIs.
--- a/app/javascript/Admin/Admin/Data/Status.elm
+++ b/app/javascript/Admin/Admin/Data/Status.elm
@@ -4,30 +4,30 @@ import Monocle.Iso exposing (..)
 
 
 type AvailabilitySatus
-    = Online
-    | Offline
+    = Active
+    | Inactive
 
 
-availablityStatusToString : AvailabilitySatus -> String
-availablityStatusToString status =
+reverseAvailablityStatusToString : AvailabilitySatus -> String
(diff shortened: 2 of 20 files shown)


tejaswinichile:
@breezyboa _a Please check.

tejaswinichile:
@breezyboa _a Please check

app/javascript/Admin/Admin/Data/Status.elm
```diff
@@ -12,10 +12,10 @@ availablityStatusToString : AvailabilitySatus -> String
 availablityStatusToString status =
     case status of
         Offline ->
-            "Offline"
+            "Online"
 
         Online ->
-            "Online"
+            "Offline"
```
breezyboa: Why are the strings reversed? Purpose of `availablityStatusToString` is to give a string representation of the status type. Reversing the strings defeats that purpose

Gemfile
```diff
@@ -113,7 +113,7 @@ end
 group :test do
   # Adds support for Capybara system testing and selenium driver
   gem "capybara", ">= 2.15", "< 4.0"
-  gem "selenium-webdriver"
+  # gem "selenium-webdriver"
```
neerajdotname: @tejaswinichile _a Please remove this entry from Gemfile rather than commenting it out.

neerajdotname:
Every reference of "offline" and "online" should be . changed to "active" and "inactive".

A document is never "offline" or "online". It's either "active" or "inactive".

For a uber driver I can see that both "active" and "inactive" as well as "online" or "offline" can work.

@tejaswinichile _a

app/javascript/Admin/Admin/Data/Status.elm
tejaswinichile: We need to show the reverse label of the status. Otherwise the current implementation we have 
 was making `online` article `online` again. 
If the article is online show label as `Make Offline` etc. I will update the functions name.

tejaswinichile:
@breezyboa _a

app/javascript/Admin/Admin/Data/Status.elm
```diff
@@ -4,30 +4,30 @@ import Monocle.Iso exposing (..)
...
     case status of
-        Offline ->
-            "Offline"
+        Inactive ->
+            "Active"
 
-        Online ->
-            "Online"
+        Active ->
+            "Inactive"
```
breezyboa: Why reverse? For an iso we need a direct mapping not an inverse

tejaswinichile: @breezyboa  check this 
https://github.com/bigbinary/acehelp/blob/master/app/javascript/Admin/Page/Article/List.elm#L176 
So according to this, If we don't reverse the string above then It will show `Mark Active` for the article which is already in `active status`, we want `Mark Inactive` lebel there. 
I checked `ISO` so to overcome this we can have one more function which will return the reverse status of current status that we will pass to `ISO` function.
Let me know your thoughts.

breezyboa: @tejaswinichile  Yes, we already have a function which returns the inverse. `statusToButtonText `.  Maybe we can rename it and make it more generic. Lets use this function for labels

tejaswinichile:
@breezyboa  _a Please have a look.

Issue #472, opened 2018-08-27
Edit article should not auto redirect to listing on save


breezyboa:
Steps
1 - Login
2 - Goto article listing select an article
3 - Select categories for the article 
4- This will trigger an autosave. Which redirects the user to the article listing

This is causing a UX bug.
We can remove the auto-redirection from edit articles page altogether

PR #474, opened 2018-08-27
Updated save article response

tejaswinichile:
1. removed article save response for article edit from admin main module
2. added success flash message div to article common view page
3. updated model after article edit
4. added success message in article model

issue #472


tejaswinichile:
@breezyboa _a Please check

Issue #467, opened 2018-08-25
Show validation errors on signup form


neerajdotname:
https://monosnap.com/file/hoWuufJ6bfSBma1tO0zPFj6ilAsiJG

I did not enter anything and just hit Submit. I did not get any error.

For the fields which are required we should have required html attributes.

PR #481, opened 2018-08-30
Adds field validations on sign up page.

Shrikant9:
Issue [#467](https://github.com/bigbinary/acehelp/issues/467)


Shrikant9:
@breezyboa _a

app/javascript/Admin/Page/Session/SignUp.elm
```diff
@@ -137,6 +145,31 @@ signUp model =
...
+            not (Field.value model.password == Field.value model.confirmPassword)
+    in
+        if emailError then
+            "Please enter a correct email."
+        else if fieldError then
+            "Please fill in all the fields."
+        else if passwordError then
+            "Given Passwords do not match."
+        else
+            ""
```
breezyboa: This function is not needed or at least can be simplified. 
You can simply compose `validEmail` function with `validateEmpty`in `initModel` itself
Same goes for `confirmPassword`. 

When you run the `Field.validate` or `Field.validateAll` function on  `Field`s you will get a `ValidationResult` and it will hold the error message if the validation fails

Check the documentation https://package.elm-lang.org/packages/bigbinary/elm-form-field/latest
for examples

app/javascript/Admin/Page/Session/SignUp.elm
```diff
@@ -137,6 +145,31 @@ signUp model =
...
+            [ model.firstName, model.email, model.password, model.confirmPassword ]
+
+        fieldError =
+            not (isAllValid fields)
+
+        passwordError =
+            not (Field.value model.password == Field.value model.confirmPassword)
+    in
+        if emailError then
+            "Please enter a correct email."
```
breezyboa: Please enter a valid email

Shrikant9:
@breezyboa _a
I made the changes for email field to be validated as a part of `isValidAll`. 
Although since `confirmPassword` requires an additional parameter i.e. `password`, a helper function that works like `validateEmail` cannot be used here, since it only passes the value of the current field.

Therefore I'm leaving the manual confirm password check in the model.

Issue #483, opened 2018-08-31
Error message needs to be better


neerajdotname:
https://monosnap.com/file/G5L6IMBy0RDSr0tsq8Gg9j6oc4ybOo

Error message needs to be "Email is already registered. Please try a different email".

@Shrikant9 _a

Shrikant9:
@tejaswinichile I get the following when I call the signup query with existing email:
```
{
  "data": {
    "signup": {
      "user": null
    }
  }
}
```

Can this be changed to an error?

tejaswinichile:
@Shrikant9 Let me check that.

PR #489, opened 2018-09-03
Fixes error message for user who is already signed up.

Shrikant9:
Issue [#483](https://github.com/bigbinary/acehelp/issues/483).


Shrikant9:
@breezyboa _a

app/javascript/Admin/Main.elm
```diff
@@ -752,8 +752,26 @@ update msg model =
                         SignUp.update suMsg currentPageModel
                 in
                     case suMsg of
-                        SignUp.SignUpResponse (Ok id) ->
-                            updateNavigation (NavigateTo Route.Login)
+                        SignUp.SignUpResponse (Ok userWithErrors) ->
+                            -- Since successful SignUpResponse msg can return errors:
```
breezyboa: This comment can be removed

app/javascript/Admin/Page/Session/SignUp.elm
```diff
@@ -111,7 +110,7 @@ update msg model =
                 else
                     ( { model | error = Just errors }, [] )
 
-        SignUpResponse (Ok user) ->
+        SignUpResponse (Ok userWithErrors) ->
             ( { model
```
breezyboa: If the server returns errors then we should not clear out all the fields. Only the password fields can be cleared out.

PR #500, opened 2018-09-05
Upgrade to Elm 0.19

ritesh404:


ritesh404:
Note: Navigation and redirecting users after an action can now be taken care of in respective modules itself. No more do we need to handle this in only in the Main module. This is thanks to the updated `Browser.application`

There is, however, an exception i.e when we need to show a User notification. This too can be taken care of but is not handled in this PR
cc: @tejaswinichile

neerajdotname:
This is a big PR and I do not want it to have merge conflicts. I see that @tejaswinichile was busy with other work today. Merging it.

PR #505, opened 2018-09-06
allow devise token auth to handle batch request

tejaswinichile:


tejaswinichile:
Merging this to check batch request behavior on staging instance.

PR #514, opened 2018-09-07
run migration in all environment

tejaswinichile:


lib/tasks/setup.rake
```diff
@@ -19,7 +19,8 @@ task :not_production do
 end
 
 desc "Sets up the project by running migration and populating sample data"
-task setup: [:environment, :not_production, "db:drop", "db:create", "db:migrate"] do
+task setup: [:environment, :not_production, "db:drop", "db:create"] do
+  system "rake db:migrate"
```
neerajdotname: Using `system` is not right in this case since `system` eats up all exceptions. Let's say that there is an issue because of which migrations failed. While running `rake setup` one would never see any error.

https://blog.bigbinary.com/2012/10/18/backtick-system-exec-in-ruby.html

neerajdotname: A better solution is how wheel is handling this issue. https://github.com/bigbinary/wheel/blob/32babf9f73db66f2abc90ac6dd5613614dd5b655/lib/tasks/setup.rake#L22

tejaswinichile: We already have [this](https://github.com/bigbinary/wheel/blob/32babf9f73db66f2abc90ac6dd5613614dd5b655/lib/tasks/setup.rake#L22) in place, but as we discussed migrations are not running properly that's I added `system rake db:migrate` .

neerajdotname: Sorry I missed that `db:migrate` was removed. I did not see that there was a horizontal scroll.

Still using `system` will not reveal error since it eats up exceptions.  Use backtick instead of system.

neerajdotname: Issue is `Procfile`. It has no release command. https://github.com/bigbinary/acehelp/blob/master/Procfile

It should be like https://github.com/bigbinary/aceinvoice-web/blob/master/Procfile

neerajdotname: Secondly in staging we do not run `rake setup`. We run `rake setup_sample_data` since we can't drop db there.

neerajdotname: Refer https://github.com/bigbinary/aceinvoice-web/blob/master/lib/tasks/setup.rake

PR #526, opened 2018-09-10
Update setup file

tejaswinichile:
1. Added to delete OrganizationUser while setup file runs
2. checked all association and dependent destroy is present


tejaswinichile:
Merging this PR, as this change is required for all later deployments.

PR #531, opened 2018-09-11
530 redirect to home page

tejaswinichile:
Issue #530


tejaswinichile:
We are using `devise`, I have removed `users_controller` and using devise session path for sign_in_path.
Because of this, we are able to use devise's `current_user` helpers which we are setting through warden in login_mutation.
The purpose of this PR is to redirect the user to articles path if he/she is already signed_in.
@neerajdotname _a

neerajdotname:
I have a busy day with other things. @prathamesh-sonpatki _a Can you review this and other Rails related PR until Vishal joins this team.

app/controllers/application_controller.rb
```diff
@@ -9,12 +9,10 @@ class ApplicationController < ActionController::Base
 
   private
     def ensure_user_is_logged_in
-      email = warden.user["email"] if warden.user
-      @resource = User.find_by(email: email)
-      if @resource
-        new_organization_path if @resource.organizations.empty?
+      if user_signed_in?
+        new_organization_path if current_user.organizations.empty?
```
prathamesh-sonpatki: What is the use of this return value?

prathamesh-sonpatki:
@tejaswinichile _A

app/controllers/application_controller.rb
tejaswinichile: We are handling same with ELM now. We don't need this anymore.

tejaswinichile:
@prathamesh-sonpatki _a Please check.

Issue #533, opened 2018-09-12
Logout option in organization creation page


tejaswinichile:
Organization creation page doesn't have logout option.

PR #537, opened 2018-09-12
Logout option for organization creation page

tejaswinichile:
Issue #533


tejaswinichile:
@ritesh404 When are we merging this upgrade PR, as right now user might get stuck into organization creation page, if he/she doesn't want to complete the organization creation flow.

ritesh404:
Merging this. since #500 will take more time for review

Issue #532, opened 2018-09-11
GraphQl query settings for an organization, save default settings for organization.


tejaswinichile:
An organization should have the default setting record to save base URL and visibility of setting page.
It will get use to edit the visibility of setting page in setting tab

tejaswinichile:
refer PR #515

prathamesh-sonpatki:
This is completed

PR #540, opened 2018-09-14
default organisation setting

tejaswinichile:
Issue #532 
1. An organization has one setting record to save base_url and visibility of organization setting page
2. Create a default setting record when an organization gets created.

### DIFF
--- a/app/graphql/mutations/organization_mutations.rb
+++ b/app/graphql/mutations/organization_mutations.rb
@@ -19,6 +19,7 @@ class Mutations::OrganizationMutations
 
         if new_org
           organization = new_org
+          Setting.where(organization_id: organization.id).first_or_create!
         else
           errors = Utils::ErrorHandler.new.detailed_error(new_org, context)
         end
--- a/app/graphql/mutations/organization_mutations.rb
+++ b/app/graphql/mutations/organization_mutations.rb
@@ -19,7 +19,6 @@ class Mutations::OrganizationMutations
 
         if new_org
           organization = new_org
-          Setting.where(organization_id: organization.id).first_or_create!
         else
           errors = Utils::ErrorHandler.new.detailed_error(new_org, context)
         end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2018_09_12_062507) do
+ActiveRecord::Schema.define(version: 2018_09_14_092210) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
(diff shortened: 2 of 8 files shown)


db/schema.rb
```diff
@@ -120,6 +120,14 @@
     t.index ["slug"], name: "index_organizations_on_slug", unique: true
   end
 
+  create_table "settings", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.string "base_url"
+    t.boolean "visibility", default: false, null: false
+    t.uuid "organization_id"
```
neerajdotname: @tejaswinichile Please make it not nullable and please add a foreign key for it.

tejaswinichile: Foreign_key is present for the same >> https://github.com/bigbinary/acehelp/pull/540/files#diff-1acd2e7e27a227829d5d14a91c863bb6R202. I will make it `null: false`.

tejaswinichile:
@prathamesh-sonpatki _a Pease check.

app/graphql/mutations/organization_mutations.rb
```diff
@@ -19,6 +19,7 @@ class Mutations::OrganizationMutations
 
         if new_org
           organization = new_org
+          Setting.where(organization_id: organization.id).first_or_create!
```
prathamesh-sonpatki: I think `after_create` callback would be better here to create related setting record for an organization.

prathamesh-sonpatki:
@tejaswinichile _A

tejaswinichile:
@prathamesh-sonpatki _a

test/graphql/mutations/organization_mutations_test.rb
```diff
@@ -64,4 +64,15 @@ class Mutations::OrganizationMutationsTest < ActiveSupport::TestCase
       })
     end
   end
+
+  test "settings present after organization create" do
```
prathamesh-sonpatki: Please rename it to `Default settings are created after an organization is created`. We need to have proper English statements in the test descriptions.

test/graphql/mutations/organization_mutations_test.rb
```diff
@@ -64,4 +64,15 @@ class Mutations::OrganizationMutationsTest < ActiveSupport::TestCase
...
+    result = AceHelp::Client.execute(@common_org_query, input: {
+      user_id: @user.id,
+      name: "New Organisation Name",
+      email: "org_general_email@gmail.com"
+    })
+    organization_id = result.data.add_organization.organization.id
+    organization = Organization.find_by!(id: organization_id)
+    assert organization.setting
```
prathamesh-sonpatki: Please add new line before this.

Basically, we should write tests like this:

```
setup

assertions
```

It becomes much more readable that way.

prathamesh-sonpatki:
@tejaswinichile _A

tejaswinichile:
@prathamesh-sonpatki _a Please check.

PR #550, opened 2018-09-18
Update setup instructions

prathamesh-sonpatki:
@neerajdotname _A


bin/setup
```diff
@@ -20,19 +20,23 @@ chdir APP_ROOT do
   system("bundle check") || system!("bundle install")
 
   # Install JavaScript dependencies if using Yarn
-  # system("bin/yarn")
+  system!("bin/yarn")
```
neerajdotname: I wasn't aware of `system!`. This is useful. I don't see it here https://ruby-doc.org/core-2.5.0/Kernel.html 

Where is it documented.

prathamesh-sonpatki: It is defined in the same file.

``` ruby
def system!(*args)
  system(*args) || abort("\n== Command #{args} failed ==")
end

```

neerajdotname: :-) Send a PR to ruby .

PR #559, opened 2018-09-18
Fix deleting all records

### LINKED ISSUE #557: Delete all record is not working in setup file
https://github.com/bigbinary/acehelp/pull/550/files#diff-be2898951d8359ace959926e45a87e4fR28 << Is not working.
And not deleting `OrganizationUser` records.

neerajdotname:
fixes #557

@tejaswinichile _a


neerajdotname:
Merging it to test. May be because data is not getting cleared I'm not able to create a new org.

PR #569, opened 2018-09-19
Session issue resolve

tejaswinichile:
1. when user logs out destroy session for all scopes
2. clear warden user while setting user through warden


tejaswinichile:
We have resolved this with devise pages.
Closing this PR.

PR #587, opened 2018-09-21
moved login page in rails view

tejaswinichile:
In this PR I have handled sign-in page only with rails view.
Links which were redirected to the sign-in page are not fixed yet.
We will test the session issue first with this change.


tejaswinichile:
Merging it to test on staging instance.

PR #588, opened 2018-09-21
moved registration and password to rails view

tejaswinichile:


tejaswinichile:
Merging this to test on staging

Issue #580, opened 2018-09-21
Invalid urls should be prevented at client side


neerajdotname:
https://www.useloom.com/share/2d7aac26381e4ceba717ac92a7f97271

@ritesh404 _a

PR #589, opened 2018-09-21
Validate Urls in Url creation

ritesh404:
#580


ritesh404:
@tejaswinichile _a can you please review

app/javascript/Admin/Page/View.elm
```diff
@@ -144,3 +144,18 @@ logoutOption signOut =
                 ]
             ]
         ]
+
+
+errorAlert : List String -> Html msg
```
neerajdotname: @ritesh404 Since this new function `errorAlert` is in generic `View.elm` page I'm assuming that it is a common helper which other pages also might want to use it.

Currently `errorAlert` is only being used in `url/create.elm`. 

Is there a plan to use this helper in other pages or will other resources will have their own way of displaying errors?

Issue #504, opened 2018-09-06
Error message is not correct


neerajdotname:
https://monosnap.com/file/P5ons80ydcyiMHDr7lolGDshH90kIL

@tejaswinichile _a

tejaswinichile:
Right now for all saving mutation, we have introduced  `Maybe object` that's why errors are not rendering at all if anything goes wrong.

tejaswinichile:
@ritesh404  I will need your help in this to handle the above condition.

PR #591, opened 2018-09-21
Handle category error messages

tejaswinichile:
Issue #504

### DIFF
--- a/app/javascript/Admin/Page/Team/Create.elm
+++ b/app/javascript/Admin/Page/Team/Create.elm
@@ -97,7 +97,7 @@ update msg model =
             ( model, [] )
 
         SaveTeamResponse (Err error) ->
-            ( { model | error = Just "An error occured while saving the Team" }, [] )
+            ( { model | error = Just "Team member already present in current organization." }, [] )
 
 
 
--- a/app/javascript/Admin/Page/Team/Create.elm
+++ b/app/javascript/Admin/Page/Team/Create.elm
@@ -97,7 +97,7 @@ update msg model =
             ( model, [] )
 
         SaveTeamResponse (Err error) ->
-            ( { model | error = Just "Team member already present in current organization." }, [] )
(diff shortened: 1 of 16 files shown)


tejaswinichile:
@ritesh404 _a

app/javascript/Admin/Page/Team/Create.elm
```diff
@@ -97,7 +97,7 @@ update msg model =
             ( model, [] )
 
         SaveTeamResponse (Err error) ->
-            ( { model | error = Just "An error occured while saving the Team" }, [] )
+            ( { model | error = Just "Team member already present in current organization." }, [] )
```
neerajdotname: I need to have a call to better understand this but how do we know that this was the issue. It could be some other issue like network or db error. Shouldn't the error message be already there in `error` ? 

Let's have a call on Monday since I want to understand how the whole thing flows. On Monday @vishaltelangre will also join the call and the project.

app/models/category.rb
```diff
@@ -12,6 +12,7 @@ class Category < ApplicationRecord
   belongs_to :organization
 
   validates :name, presence: true
+  # validates_uniqueness_of :name, case_sensitive: false
```
neerajdotname: @tejaswinichile _a remove this line rathe than commenting it out.

neerajdotname:
If it is a generic error then we should keep the message that  was earlier. Here we are changing to a specific message which might or might not be true.

app/models/category.rb
tejaswinichile: We should have this validation, but we need to fix the setup file according to it.

app/javascript/Admin/Page/Team/Create.elm
tejaswinichile: `NetworkError` `BadPayloadError` all are included in `SaveTeamResponse (Ok response)` this particular error is handled by `SaveTeamResponse (Err error)`,  that's why I have updated this text.

PR #602, opened 2018-09-24
Fix: Various issues while initial set up of the app in development environment

vishaltelangre:


bin/setup
```diff
@@ -29,7 +35,7 @@ chdir APP_ROOT do
   end
 
   puts "\n== Setting up the app =="
-  system! "bin/rake setup"
+  system! "bundle exec rake setup"
```
vishaltelangre: It appears that the logic inside the initializer  `initializers/schedule_recurring_job_for_first_time.rb` is getting invoked while the `setup` rake task is executed.

During my first attempt, the command `./bin/setup` was failed with the following error.

```
[...]
== Setting up the app ==
WARNING: Nokogiri was built against LibXML version 2.6.30, but has dynamically loaded 2.9.4
/Users/vishaltelangre/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/activerecord-5.2.1/lib/active_record/connection_adapters/postgresql_adapter.rb:688:in `rescue in connect': FATAL:  database "acehelp_development" does not exist
 (ActiveRecord::NoDatabaseError)
[...]
	from /Users/vishaltelangre/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/delayed_job_recurring-0.3.7/lib/delayed/recurring_job.rb:191:in `schedule!'
	from /Users/vishaltelangre/projects/acehelp/config/initializers/schedule_recurring_job_for_first_time.rb:2:in `<main>'
[...]
== Command ["bin/rake setup"] failed ==
[...]
```

To fix the above error, I manually created `acehelp_development` database using the command `createdb acehelp_development`.

After fixing the above error, now re-running the `./bin/setup` command failed with the following error.

```
[...]
== Setting up the app ==
WARNING: Nokogiri was built against LibXML version 2.6.30, but has dynamically loaded 2.9.4
/Users/vishaltelangre/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/activerecord-5.2.1/lib/active_record/connection_adapters/postgresql_adapter.rb:603:in `async_exec': PG::UndefinedTable: ERROR:  relation "delayed_jobs" does not exist
LINE 1: SELECT "delayed_jobs".* FROM "delayed_jobs" WHERE (((handler...
                                     ^
: SELECT "delayed_jobs".* FROM "delayed_jobs" WHERE (((handler LIKE '--- !ruby/object:AutoCloseResolvedTicketsJob %') OR (handler LIKE '--- !ruby/object:AutoCloseResolvedTicketsJob
%'))) (ActiveRecord::StatementInvalid)
[...]
	from /Users/vishaltelangre/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/delayed_job_recurring-0.3.7/lib/delayed/recurring_job.rb:191:in `schedule!'
	from /Users/vishaltelangre/projects/acehelp/config/initializers/schedule_recurring_job_for_first_time.rb:2:in `<main>'
[...]
```

From the above error, the same `config/initializers/schedule_recurring_job_for_first_time.rb` initializer seems to be the issue which was preventing the local app setup.

What I have found is that when the `bin/rake setup` command is run, it starts the `spring` program instead of `rake` command.

```
$ ps aux | grep rake
$ ps aux | grep spring
vishaltelangre   76721   0.0  0.1  4315104   9828 s002  S     8:13AM   0:00.57 spring server | acehelp | started 1 secs ago
```

Due to this, the global variable `$PROGRAM_NAME` used in https://github.com/bigbinary/acehelp/blob/72c8b43f9fb2954a176f55b51ed8ee43e416ce03/config/initializers/schedule_recurring_job_for_first_time.rb#L1 makes the condition of the if block as `true` which results in both of the above-mentioned errors.

I didn't want to touch that initializer and just wanted to run the setup task successfully.
Instead, I changed the `bin/rake setup` command to `bundle exec rake setup` which worked successfully on a freshly cloned repository.

bin/setup
```diff
@@ -15,6 +15,12 @@ chdir APP_ROOT do
   # This script is a starting point to setup your application.
   # Add necessary setup steps to this file.
 
+  puts "== Check external dependencies =="
+  puts "Check whether elasticsearch is installed or not"
+  system! "elasticsearch --version"
+  puts "Check whether elasticsearch service is running on 9200 port or not"
+  system! "curl -s -S localhost:9200 > /dev/null"
+
```
vishaltelangre: The command `./bin/setup` was failing for me with the following error.

```
[...]
== Reindex searchkick ==
WARNING: Nokogiri was built against LibXML version 2.6.30, but has dynamically loaded 2.9.4
Reindexing Article...
rake aborted!
Faraday::ConnectionFailed: Failed to open TCP connection to localhost:9200 (Connection refused - connect(2) for "localhost" port 9200)
[...]
```

Above checks ensures that the dependency `elasticsearch` is installed and running before further execution.

vishaltelangre:
@neerajdotname _a Please review.

PR #599, opened 2018-09-24
Add cancel button to create article page

vishaltelangre:
https://monosnap.com/file/aTl2pqd2XVDwB543cdRBtScYmmILD0

@vishaltelangre _a

### DIFF
--- a/app/javascript/Admin/Page/Article/Create.elm
+++ b/app/javascript/Admin/Page/Article/Create.elm
@@ -69,6 +69,7 @@ type Msg
     | CategorySelected (List CategoryId)
     | UrlsLoaded (Result GQLClient.Error (Maybe (List UrlData)))
     | UrlSelected (List UrlId)
+    | Cancel
 
 
 update : Msg -> Model -> ( Model, List (ReaderCmd Msg) )
@@ -132,6 +133,9 @@ update msg model =
             , []
             )
 
+        Cancel ->
+            ( model, [] )
+
 
 
 -- View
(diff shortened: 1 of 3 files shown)


vishaltelangre:
@neerajdotname _a Please review.

Right now, this feature does not ask for the confirmation from the user whether to leave the page or not when the form is changed. I am not sure how do we display a confirmation to dialog to the user.

neerajdotname:
Right now the whole application does not have any confirmation. We will deal with confirmation at the whole project level.

@ritesh404 _a Please review.

app/javascript/Admin/Page/Article/Create.elm
```diff
@@ -167,6 +171,9 @@ view model =
                 [ multiSelectCategoryList "Categories:" model.categories CategorySelected
                 , multiSelectUrlList "Urls:" model.urls UrlSelected
                 , button [ id "create-article", type_ "button", class "btn btn-success", onClick SaveArticle ] [ text "Create Article" ]
+                , button
```
ritesh404: @vishaltelangre we can change the type signature of the view function to
```
view : ApiKey -> Model -> Html Msg
```

Then instead of creating a new `Cancel` Msg constructor  and have a button with onclick we can simply use the anchor tag with href  to ` routeToString <| ArticleList orgKey`

Thus the routing mechanism will automatically kick in and will take care of the rest

Have a look at this for reference
https://github.com/bigbinary/acehelp/blob/bf641e5f3ba439cba0ffd4ef1e7f1629e650e82e/app/javascript/Admin/Page/Article/List.elm#L156

vishaltelangre:
@ritesh404 _a Incorporated the feedback. Please review. 

The preview app on Heroku still shows cancel as a button instead of a link.

PR #621, opened 2018-09-25
added test case to check visibility is true by default

tejaswinichile:
Issue #613


tejaswinichile:
Not a big change here merging this PR.

Issue #608, opened 2018-09-24
Cannot create an article if content is pasted in from clipboard


ritesh404:
https://monosnap.com/file/xGYPMGtGUaBhlzkp0BHupXYInjSgeb

PR #622, opened 2018-09-25
Fix: Cannot create an article if content is pasted in from clipboard

vishaltelangre:
Fixes https://github.com/bigbinary/acehelp/issues/608.


vishaltelangre:
@ritesh404 _a Please review.

Issue #617, opened 2018-09-25
Better UI for Ticket tab 


ritesh404:
Reference:

<img width="526" alt="tickets-ui" src="https://user-images.githubusercontent.com/7039911/45996149-bd76e000-c0b8-11e8-958f-69ca4fda609e.png">

Note the highlighted row on mouse over
Clicking on any row should take the user to edit ticket

PR #629, opened 2018-09-26
Fix UI of the "Ticket" tab

vishaltelangre:
Fixes https://github.com/bigbinary/acehelp/issues/617.

### DIFF
--- a/app/javascript/Admin/Page/Ticket/List.elm
+++ b/app/javascript/Admin/Page/Ticket/List.elm
@@ -101,7 +101,7 @@ rows : ApiKey -> Model -> Ticket -> Html Msg
 rows orgKey model ticket =
     tr
         []
-        [ td [ class "row-id" ] [ text ticket.id ]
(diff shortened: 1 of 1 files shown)


vishaltelangre:
Screenshot:

image

@ritesh404 _a Please review.

I have used the table markup provided by Twitter Bootstrap 4.0 to design the UI of this page. Enclosing a `tr` tag inside an `a` tag was disturbing the whole markup and is also not a recommended approach. For now, I have added an additional column which displays "View" link for every row in the table.

If we still want to make the whole row clickable, instead of using an `a` tag, we will need to use the Router to update the navigation programatically in the Main module by attaching an `onClick` handler on every `tr` node. Please let me know your thoughts.

neerajdotname:
> If we still want to make the whole row clickable

I don't think that is needed. In fact in many cases ability to click the whole row has caused me to click on things I did not mean to click. This is good enough for now.

Issue #628, opened 2018-09-26
Trix editor should be more friendly to type at the bottom


neerajdotname:
One can experiment with how basecamp handles such cases.

https://www.useloom.com/share/ae53d8a980b843e2bcdc1d4ae0b51113

vishaltelangre:
@neerajdotname @ritesh404 

The height of the Trix editor integrated into Basecamp dynamic, that is it depends on the size of the contents inside the editor.

I think that's a better default than trying to fit the whole editor into the visible viewport having the fixed width-height dimensions. With the default settings, the editor does not have its own scrollbar since it expands the page itself vertically with the contents inside it.

I am going to assign this to myself and work adding some space between the editor and the bottom of the page to reduce the visual confusion.

If it doesn't work out as expected, we will try to display the editor into a container with the fixed dimensions and also with vertical scrollbars of its own.

@vishaltelangre _a

neerajdotname:
https://www.useloom.com/share/5d3f4e6ae0004a1e99177d97cabac3ba

vishaltelangre:
Fixed in #631.

PR #631, opened 2018-09-26
Trix editor should be more friendly to type at the bottom

vishaltelangre:
Possible fix for https://github.com/bigbinary/acehelp/issues/628.


vishaltelangre:
I agree with you but I am not sure how that can be done though. Can we
calculate these things in Elm?

ritesh404:
@vishaltelangre I believe we can. You can have a look at the `Browser.Dom`
https://package.elm-lang.org/packages/elm/browser/latest/Browser-Dom

vishaltelangre:
@ritesh404 Thanks. I will take a look at it.

vishaltelangre:
@vishaltelangre _a

vishaltelangre:
@ritesh404 _a This is ready now. Please review.

- This feature uses Browser.Dom package to make the calculations. 
- A JS port is used just for setting the style attribute. 
- Also, this solution is window resize-aware. 
- Additionally, the editor is adjusted when the non-editable content is made editable by clicking on the "Edit" button.

Issue #625, opened 2018-09-26
Error messages are not appearing correctly


neerajdotname:
https://www.useloom.com/share/c3b9ee11cb8048f4ba422c967ebe4857

PR #632, opened 2018-09-26
render error message for registration page

tejaswinichile:
Issue #625


tejaswinichile:
No major change here merging it.

Issue #605, opened 2018-09-24
Flash of unstyled content on sign up and sign in pages


ritesh404:
<iframe width="560" height="315" src="https://monosnap.com/video/IV5bnDGfO3AHTKb2BIrCBtbbmyJDMe" frameborder="0" controls allowfullscreen></iframe>

tejaswinichile:
The style was taking time to load, I have added one solution for the same.
And yeah it works same with `stylesheet_link_tag` too with precompiled assets but giving the same problem so have to include it before the form renders.

cc @ritesh404 @prathamesh-sonpatki

PR #635, opened 2018-09-26
load styling before rendering the form

tejaswinichile:
#605


tejaswinichile:
@prathamesh-sonpatki _a

Issue #627, opened 2018-09-26
Article show page should have article highlighted


neerajdotname:
https://monosnap.com/file/8vmjyiGqWQmuIpS4kJUx1Enz3npZM1

ritesh404:
This is happening in the following areas also
> Add new team member
> View ticket
> Edit Url
> Edit Category

PR #637, opened 2018-09-27
Make nav link active if the current page's route matches a substring

vishaltelangre:
Fixes https://github.com/bigbinary/acehelp/issues/627.


vishaltelangre:
@ritesh404 _a Please review.

Issue #638, opened 2018-09-27
update setup file and api-key for careforever and ace-invoice


tejaswinichile:

tejaswinichile:
I have done it for care-forever site, able to fetch data according to url.

PR #640, opened 2018-09-27
Update setup for careforever

tejaswinichile:
#638


tejaswinichile:
Merging it to set up on careforever site.

PR #641, opened 2018-09-27
prevent data reset after every deployment

tejaswinichile:
This change is necessary to keep API-Key constant after every deployment.


neerajdotname:
I don't approved of it. In the long run it will cause issues.

I had mentioned that we should hardcode the api key.

tejaswinichile:
@neerajdotname You meant while saving the data. Okay, will do that. 👍

PR #644, opened 2018-09-27
upload end point for article image upload

tejaswinichile:
1. added method to upload pic in article controller
2. updated article update mutation to update article persist flag to true
3. added column to maintain persist flag with default false in articles table
4. initialized active_storage attachments for article

### DIFF
--- a/app/controllers/articles/attachments_controller.rb
+++ b/app/controllers/articles/attachments_controller.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+class Articles::AttachmentsController < ApplicationController
+  before_action :ensure_user_is_logged_in
+  include LoadOrganization
+  before_action :load_article, only: :create
+
+  def create
+    attachments = @article.attachments.attach(
+      io: attachment_params[:file],
+      content_type: attachment_params[:content_type],
+      filename: attachment_params[:filename]
+    )
+    render json: { url: url_for(attachments.last) }, status: :created
+  end
+
+  private
+
+    def attachment_params
+      params.require(:attachment).permit(:file, :filename, :content_type)
+    end
+
+    def load_article
+      @article = Article.find(params[:id])
+    end
+end
--- a/app/controllers/articles_controller.rb
+++ b/app/controllers/articles_controller.rb
@@ -3,8 +3,23 @@
 class ArticlesController < ApplicationController
   before_action :ensure_user_is_logged_in
   include LoadOrganization
+  before_action :article, only: :upload
 
   def index
     render
   end
+
+  def upload
+    @article.images.attach(article_params[:images])
+  end
+
+  private
+
+    def article_params
+      params.require(:article).permit(images: [])
+    end
+
+    def article
+      @article = Article.find_by!(id: params[:id])
+    end
 end
--- a/app/controllers/articles_controller.rb
+++ b/app/controllers/articles_controller.rb
@@ -3,23 +3,8 @@
 class ArticlesController < ApplicationController
   before_action :ensure_user_is_logged_in
   include LoadOrganization
-  before_action :article, only: :upload
 
   def index
     render
   end
-
-  def upload
-    @article.images.attach(article_params[:images])
-  end
-
-  private
-
-    def article_params
-      params.require(:article).permit(images: [])
-    end
-
-    def article
-      @article = Article.find_by!(id: params[:id])
-    end
 end
--- a/app/graphql/mutations/article_mutations.rb
+++ b/app/graphql/mutations/article_mutations.rb
@@ -59,7 +59,7 @@ class Mutations::ArticleMutations
       if article.nil?
         errors = Utils::ErrorHandler.new.error("Article not found", context)
       else
-        if article.update_attributes(title: inputs[:title], desc: inputs[:desc], persist: true)
+        if article.update_attributes(title: inputs[:title], desc: inputs[:desc], temporary: false)
           if inputs[:category_ids].present?
             article.category_ids = inputs[:category_ids]
           end
--- a/app/graphql/mutations/article_mutations.rb
+++ b/app/graphql/mutations/article_mutations.rb
@@ -59,7 +59,7 @@ class Mutations::ArticleMutations
       if article.nil?
         errors = Utils::ErrorHandler.new.error("Article not found", context)
       else
-        if article.update_attributes(title: inputs[:title], desc: inputs[:desc])
+        if article.update_attributes(title: inputs[:title], desc: inputs[:desc], persist: true)
           if inputs[:category_ids].present?
             article.category_ids = inputs[:category_ids]
           end
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -14,6 +14,7 @@ class Article < ApplicationRecord
 
   has_many :article_categories, dependent: :destroy
   has_many :categories, through: :article_categories
+  has_many_attached :images
 
   validates :title, uniqueness: { scope: [:organization_id] }, presence: true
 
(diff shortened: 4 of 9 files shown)


tejaswinichile:
@vishaltelangre I have added the endpoint with the path :
`/organizations/:organization_api_key/articles/:id/upload`
Please test it with article edit, till I modify article elm module to load article for article new page.

app/controllers/articles_controller.rb
```diff
@@ -3,8 +3,23 @@
...
   include LoadOrganization
+  before_action :article, only: :upload
 
   def index
     render
   end
+
+  def upload
+    @article.images.attach(article_params[:images])
+  end
```
vishaltelangre: We can extract this non-RESTful custom action into a RESTful controller. For example, an Articles::UploadsController with a #create action. 

Reference: http://jeromedalbert.com/how-dhh-organizes-his-rails-controllers/.

vishaltelangre: Attachements sounds a better name insteaf of Uploads or Images to me.

app/controllers/articles_controller.rb
```diff
@@ -3,8 +3,23 @@
...
+
+  private
+
+    def article_params
+      params.require(:article).permit(images: [])
+    end
+
+    def article
+      @article = Article.find_by!(id: params[:id])
```
vishaltelangre: Please rename this method as #load_article.

Also, `Article.find_by!(id: params[:id])` can be simply written as `Article.find(params[:id])`.

app/graphql/mutations/article_mutations.rb
```diff
@@ -59,7 +59,7 @@ class Mutations::ArticleMutations
       if article.nil?
         errors = Utils::ErrorHandler.new.error("Article not found", context)
       else
-        if article.update_attributes(title: inputs[:title], desc: inputs[:desc])
+        if article.update_attributes(title: inputs[:title], desc: inputs[:desc], persist: true)
```
vishaltelangre: `persisted` sounds okay to me, other alternatives are `initial`, `temporary` etc.

app/models/article.rb
```diff
@@ -14,6 +14,7 @@ class Article < ApplicationRecord
 
   has_many :article_categories, dependent: :destroy
   has_many :categories, through: :article_categories
+  has_many_attached :images
```
vishaltelangre: We may need to support other types of attachments, so instead of :images, we could say :attachments.

vishaltelangre:
@tejaswinichile Can you please make the endpoint URL, supported requests parameters and the response JSON as same as specified in here - https://3.basecamp.com/3091787/buckets/9258656/todos/1326409399#__recording_1326556541 ?

app/graphql/mutations/article_mutations.rb
tejaswinichile: `persisted` gives attribute errors as it's defined by active records, `temporary` will work here.

vishaltelangre: Idea is to have a meaningful name.

tejaswinichile:
@vishaltelangre I have made mentioned changes here.
You can hit attachment request to :
`POST   /organizations/:organization_api_key/articles/:article_id/attachments`

app/controllers/articles/attachments_controller.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class Articles::AttachmentsController < ApplicationController
+  before_action :ensure_user_is_logged_in
+  include LoadOrganization
+  before_action :load_article, only: :create
+
+  def create
+    @article.attachments.attach(params[:attachments])
```
vishaltelangre: We will not pass bulk attachments at once. Please accept only a single attachment here.

Also, please use strong parameters to require and whitelist the params as follows.

```ruby
def attachment_params
  params.require(:attachment).permit(:file, :filename, :content_type)
end
```

app/controllers/articles/attachments_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    if @article.attachments.attached?
+      render json: { url:  url_for(@article.attachments.last) }, status: :created
```
vishaltelangre: Remove additional space between the `url: url_for(` here.

vishaltelangre: The key name should be `attachment_url` instead of `url`. Also, include `succees: true` for success branch, `success: false` in the error branch.

config/routes.rb
```diff
@@ -16,9 +16,7 @@
 
   resources :organizations, only: [:show], param: :api_key do
     resources :articles, only: [:index] do
-      member do
-        put :upload
-      end
+      resource :attachments, only: [:create]
```
vishaltelangre: `only: :create` works as well

vishaltelangre:
@tejaswinichile _a Please take care of the recent comments as well.

app/controllers/articles/attachments_controller.rb
tejaswinichile: @vishaltelangre I didn't add it because I didn't know how you are going to send the params.
Will it be inside the article or an individual param hash.
Anyways you can always change it according to your request.

vishaltelangre: @tejaswinichile Since article ID and organization ID will be already present in the endpoint URL, the upload request will have the following multi-part/form data which is already specified here - https://3.basecamp.com/3091787/buckets/9258656/todos/1326409399#__recording_1326556541.

```
attachment: {
  filename: "file_name_here.png",
  content_type: "image/png",
  file: ActionDispatch::Http::UploadedFile<...>
}
```

app/controllers/articles/attachments_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+  before_action :ensure_user_is_logged_in
+  include LoadOrganization
+  before_action :load_article, only: :create
+
+  def create
+    @article.attachments.attach(
+      file: attachment_params[:file],
+      content_type: attachment_params[:attachment_params],
+      filename: attachment_params[:attachment_params]
+    )
```
vishaltelangre: This r[eturns an array](https://api.rubyonrails.org/classes/ActiveStorage/Attached/Many.html#method-i-attach) with a single `ActiveStorage::Attachment` record which was attached. You can use that for the later use instead of using `@article.attachments.last` which could result in ambiguous situations in certain race conditions.

app/controllers/articles/attachments_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+    if @article.attachments.attached?
```
vishaltelangre: Can `@article.attachments.attached?` be true if there already exist some attachments before the current one? In that case, this condition doesn't seem right.

app/controllers/articles/attachments_controller.rb
vishaltelangre: Also, according to the docs, it should be `io` key instead of `file` provided to the #attach method.

vishaltelangre: Typos:

- `content_type: attachment_params[:attachment_params]` should be `content_type: attachment_params[:content_type],`
- `filename: attachment_params[:attachment_params]` should be `filename: attachment_params[:filename]`

app/controllers/articles/attachments_controller.rb
tejaswinichile: It was making sense in bulk upload. Removing this check now.

tejaswinichile:
@vishaltelangre Please have a look. And merge it if you feel it's okay.

vishaltelangre:
Merging it. I will rebase and make required changes further.

PR #648, opened 2018-09-28
article loading while creating new one

tejaswinichile:
1. updated article mutations to updated article
2. load article while creating new one
3. added temporary article query
4. updated article page and data request module according to it

### DIFF
--- a/app/javascript/Admin/Page/Article/Create.elm
+++ b/app/javascript/Admin/Page/Article/Create.elm
@@ -1,10 +1,20 @@
-module Page.Article.Create exposing (Model, Msg(..), articleInputs, init, initModel, save, update, view)
+module Page.Article.Create exposing
+    ( Model
+    , Msg(..)
+    , articleInputs
+    , init
+    , initModel
+    , save
+    , update
+    , view
+    )
 
 import Admin.Data.Article exposing (..)
 import Admin.Data.Category exposing (..)
 import Admin.Data.Common exposing (..)
 import Admin.Data.ReaderCmd exposing (..)
 import Admin.Data.Url exposing (UrlData, UrlId)
+import Admin.Ports exposing (..)
 import Admin.Request.Article exposing (..)
 import Admin.Request.Category exposing (..)
 import Admin.Request.Helper exposing (ApiKey)
@@ -28,32 +38,35 @@ import Task exposing (Task)
 
 
 type alias Model =
-    { title : Field String String
+    { articleId : ArticleId
+    , title : Field String String
     , desc : Field String String
-    , articleId : Maybe ArticleId
     , categories : List (Option Category)
     , urls : List (Option UrlData)
     , status : SaveStatus
     , errors : List String
+    , success : Maybe String
     }
 
 
 initModel : Model
 initModel =
-    { title = Field (validateEmpty "Title") ""
+    { articleId = ""
+    , title = Field (validateEmpty "Title") ""
     , desc = Field (validateEmpty "Article Content") ""
-    , articleId = Nothing
     , categories = []
     , urls = []
     , status = None
     , errors = []
+    , success = Nothing
     }
 
 
 init : ( Model, List (ReaderCmd Msg) )
 init =
     ( initModel
-    , [ Strict <| Reader.map (Task.attempt CategoriesLoaded) requestCategories
+    , [ Strict <| Reader.map (Task.attempt ArticleLoaded) requestTemporaryArticle
+      , Strict <| Reader.map (Task.attempt CategoriesLoaded) requestCategories
       , Strict <| Reader.map (Task.attempt UrlsLoaded) requestUrls
       ]
     )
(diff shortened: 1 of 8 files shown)


app/javascript/Admin/Page/Article/Create.elm
```diff
@@ -86,17 +100,36 @@ update msg model =
...
+                      }
+                    , []
+                    )
+
+        SaveArticleResponse (Err error) ->
             ( { model
-                | title = Field.update model.title ""
-                , desc = Field.update model.desc ""
+                | errors = [ "There was an error while saving the article" ]
+                , status = None
```
vishaltelangre: I think both the branches such as ` SaveArticleResponse (Ok articleResp)` and `SaveArticleResponse (Err error)` can be made reusable since the similar branches exist in the `Article.Edit` module as well.

vishaltelangre:
@ritesh404 _a Please review this? I have left just one comment. Other sutff looks okay to me.

app/javascript/Admin/Page/Article/Create.elm
tejaswinichile: Both branches have the same response but not the data, we don't have `status, isEditable` checks in article create page.
And in another way, I was thinking to redirect page to article edit page after successful save.

vishaltelangre: We can provide the consistent response from the backend to both of the pages. Then, on the new page, we can ignore information such as status which is not needed.

tejaswinichile: That's what we are doing it right now, we are ingnoring status here, so how can we extract these things if they are not same.

tejaswinichile:
@vishaltelangre Yeah, we have not null constraint on title and description.
Anyways we will not be showing canceled request in listing remember, that's still in my to-do.

app/javascript/Admin/Page/Article/Create.elm
vishaltelangre: We can have the same alias type for both the new and edit pages. Therefore, so many things can be shared. Only view logic would be different. For example, view of the new page wouldn't display the status although its model has it, whereas the edit page would display it.

ritesh404:
Articles with the temporary flag as true should not be sent to the client in the `articles` query

app/javascript/Admin/Page/Article/Create.elm
```diff
@@ -28,32 +38,35 @@ import Task exposing (Task)
...
     , status : SaveStatus
     , errors : List String
+    , success : Maybe String
     }
 
 
 initModel : Model
 initModel =
-    { title = Field (validateEmpty "Title") ""
+    { articleId = ""
```
ritesh404: This can still be represented by a Maybe type. If we have a Nothing. that means Either we are waiting for a response from the temporary article create query. Or something went wrong.
Maybe would be still better here than having an empty string

app/javascript/Admin/Page/Article/Create.elm
ritesh404: Article status by default should be Inactive. If we can unify this and a few more things. Probably we can have one single page to handle Edit and Create.

app/javascript/Admin/Page/Article/Create.elm
```diff
@@ -127,6 +160,26 @@ update msg model =
...
+                    ( { model
+                        | articleId = article.id
+                        , errors = []
+                      }
+                    , []
+                    )
+
+                Nothing ->
+                    ( { model
+                        | errors = [ "There was an error loading the article" ]
```
ritesh404: This would seem odd. The User is trying to create a new article. If for some reason something goes wrong and a temporary article is not created this error would show up and it would be quite confusing for the user.

I am not sure what would be the right thing to do here. As showing just this message would not help. As the user can still save an article. We might also need to disable the save button. And maybe the message can be "Something went wrong could you please try again later?" 

cc: @vishaltelangre

app/javascript/Admin/Page/Article/Create.elm
```diff
@@ -86,17 +100,36 @@ update msg model =
         SaveArticle ->
             save model
```
ritesh404: Before saving we need to check if we have a valid articleId. If we use the Maybe type here it will help

app/javascript/Admin/Page/Article/Create.elm
tejaswinichile: I will be removing this for now.

tejaswinichile:
https://github.com/bigbinary/acehelp/pull/648#discussion_r221222020 
<< Here status is nothing but saving status.

app/javascript/Admin/Page/Article/Create.elm
tejaswinichile: Here status is nothing but saving status.
@ritesh404 I am not sure how we will move this to the common page to implement the update function for saveArticleResponse.

tejaswinichile:
@ritesh404 _a Please have a look.

app/javascript/Admin/Page/Article/Create.elm
vishaltelangre: Ideally, an article's status should not be `None`, instead, it should be strictly either `Active` or `Inactive`.

neerajdotname:
Agree about the article status.

app/javascript/Admin/Page/Article/Create.elm
tejaswinichile: It's not article status. We have articleStatus as a different entity, here you are confusing with article saving status.
You can refer the article edit page to get an idea.

vishaltelangre: Yes, I got confused. Looks like this status is of type `SaveStatus` which is different from the type `AvailabilityStatus`. Maybe we should rename that property as `saveStatus` in the model to reduce the confusion.

ritesh404: Yes lets rename it to `saveStatus`. @tejaswinichile _a can you please do this.
Rest looks good

tejaswinichile:
@ritesh404 _a Please check. I have made changes for `saveStatus` for both create and edit pages.

Issue #563, opened 2018-09-19
Show hamburger menu so that user can create more organizations just like in AceInvoice


neerajdotname:
https://monosnap.com/file/XAAitkpNFtKFloZNs5HXSS0ziMctUD

PR #649, opened 2018-09-28
Hamburger Menu

ritesh404:
#563


ritesh404:
@tejaswinichile _a can you please review the elm code

ritesh404:
@ritesh404 _a assigning it back to me to add the "Add New Organization" button

ritesh404:
@tejaswinichile _a please review

tejaswinichile:
When we change the organization from the hamburger and try to reload then the default organization gets loaded.
But in case of when we change the organization then do some action and the try to reload then it works fine.
finding: when we switch the organization URL is not updating right away.
@ritesh404 _a Please check.

ritesh404:
@tejaswinichile _a please check now. Also please rebase with master

PR #654, opened 2018-10-01
render only persisted records in article list

tejaswinichile:


tejaswinichile:
@vishaltelangre _a Please check

app/models/article.rb
```diff
@@ -31,6 +31,7 @@ def increment_downvote
   def self.search_using(org, opts = {})
     articles = opts[:status].present? ? Article.send(opts[:status]) : Article.all
     articles = articles.for_organization(org)
+    articles = articles.where(temporary: false)
```
vishaltelangre: @tejaswinichile _a This `Article::search_using` method is doing so many things. Please extract it into small chainable named scopes. Ideally, the number of lines in a method should not exceed 10.

Rest looks good.

tejaswinichile: Cool 👍

tejaswinichile:

app/models/article.rb
```diff
@@ -20,6 +20,18 @@ class Article < ApplicationRecord
 
   scope :for_organization, ->(org) { where(organization: org) }
 
+  scope :persisted_articles, -> { where(temporary: false) }
+
+  scope :search_with_id, ->(opts) { opts[:article_id] && where(id: opts[:article_id]) }
```
vishaltelangre: This scope seems unnecessary. The same thing can be achieved using `where(id: article_id_here)`.

app/models/article.rb
```diff
@@ -20,6 +20,18 @@ class Article < ApplicationRecord
...
+
+  scope :search_with_status, ->(opts) { opts[:status] && Article.send(opts[:status]) }
```
vishaltelangre: Why do we need a hash as an argument here?  Also, `Article.send` isn't a good practice.

Probably, `sope :with_status -> (status) { where(status: status) }` would do the job?

app/models/article.rb
```diff
@@ -20,6 +20,18 @@ class Article < ApplicationRecord
...
+
+  scope :search_with_url, ->(opts) {
+    opts[:url] && joins(:urls).where(
+      "urls.url = ?", opts[:url]
+    )
+  }
```
vishaltelangre: Same as above, no need to accept a hash (opts) here.

app/models/article.rb
```diff
@@ -29,14 +41,9 @@ def increment_downvote
...
   def self.search_using(org, opts = {})
-    articles = opts[:status].present? ? Article.send(opts[:status]) : Article.all
-    articles = articles.for_organization(org)
-    articles = articles.where(id: opts[:article_id]) if opts[:article_id].present?
-    if opts[:url].present?
-      articles = articles.joins(:urls).where(
-        "urls.url = ?", opts[:url]
-      )
-    end
+    articles = Article.persisted_articles.for_organization(org)
```
vishaltelangre: Please expand `opts` to `options` for better readability.

app/models/article.rb
tejaswinichile: Yeah, It would, but we have enum declared for `status` that's why used those.

neerajdotname: Even when you have to use it , which is not the case here,  first try `public_send` and then only resort to `send`.

app/models/article.rb
tejaswinichile: @vishaltelangre But here we need one more check of the `options[:article_id]` present. In that case, we need this scope right?

tejaswinichile:
@vishaltelangre _a Please have a look.

app/models/article.rb
```diff
@@ -22,13 +22,13 @@ class Article < ApplicationRecord
...
 
-  scope :search_with_status, ->(opts) { opts[:status] && Article.send(opts[:status]) }
+  scope :search_with_status, ->(status) { status && where(status: status) }
 
-  scope :search_with_url, ->(opts) {
-    opts[:url] && joins(:urls).where(
-      "urls.url = ?", opts[:url]
+  scope :search_with_url, ->(url) {
+    url && joins(:urls).where(
+      "urls.url = ?", url
```
vishaltelangre: If this is going to be a multiline scope, please define a class method with the same name instead.

Usually, scopes are defined in just a single line.

vishaltelangre:
@tejaswinichile Please add tests for the newly introduced scopes. Also, please add tests for the `Article::search_using` method if there do not exist any.

app/models/article.rb
vishaltelangre: We can. In that case, instead of curly braces, the preferred way is to enclose the logic inside a `do..end` block when the logic is spread over multiple lines.

On the other hand, I think we can simply write it in a single line here.

tejaswinichile:
@vishaltelangre We have tests added for `Article::search_using`

app/models/article.rb
tejaswinichile: Then it exceeds the 80 columns per line. :|

Issue #658, opened 2018-10-03
Validate mime type and size of the file selected for upload


vishaltelangre:
To begin with, we need to add support for uploading images of type PNG, JPEG, and GIF.

The selected file should not exceed 5 MiB in size for the file types PNG and JPEG respectively. Whereas, files of type GIF should not exceed 10 MiB in size.

PR #659, opened 2018-10-03
Validate mime type and size of the file selected for upload

vishaltelangre:
Fixes https://github.com/bigbinary/acehelp/issues/658.


vishaltelangre:
@ritesh404 _a Please review.

Also, can we change the `prettier --tab-width 4 --write` command to indent using just 2 soft spaces?

ritesh404:
> Also, can we change the prettier --tab-width 4 --write command to indent using just 2 soft spaces?

@vishaltelangre we could. I made it 4 because in my opinion it makes code a bit more clearer. 
Although 2 I think is the default.

vishaltelangre:
We can change the indentation width in a separate PR since it will modify the files which are irrelevant to the changes made in this PR.

Issue #636, opened 2018-09-26
graphql endpoint to upload pic from editor


tejaswinichile:
related to: #626 
We need to have graphql client uploaded from elm side.

1. get file content in the base64 format, get content-type, get the filename
2. send this encoded JSON to graphql server
3. have custom type to handle file input

tejaswinichile:
- [ ] Add column to save data persisted flag
- [ ] ArticleLoad before callback in article new page with persisted status false
- [ ] Add method to upload the file with article-id and file params
- [ ] Update Article/Create mutation with article update call, we are updating article with the persisted flag false
- [ ] Update article list with filter : `persisted: true`

tejaswinichile:
As the last PR is up for the same.
Closing this issue.

PR #663, opened 2018-10-03
Add rake task to delete articles with temporary flag

tejaswinichile:
Issue #636


vishaltelangre:
@tejaswinichile _a Can you please try to wrap this up before moving on to other tasks?

Also, please create another issue to make a delete request to purge the temporary article whenever user tries to navigate away without saving the article on the create article page.

tejaswinichile:
@vishaltelangre _a You can review it now.

Gemfile
```diff
@@ -78,6 +78,12 @@ gem 'browser'
 # for error tracking
 gem 'honeybadger', '~> 4.0'
 
+# For background processing
+gem 'sidekiq'
+
+# For periodic jobs on sidekiq
+gem 'sidekiq-cron'
```
vishaltelangre: Is there any issues with the https://github.com/amitree/delayed_job_recurring gem which we are using right now? We should not be using both delayed_job and sidekiq at the same.

app/models/article.rb
```diff
@@ -15,6 +15,7 @@ class Article < ApplicationRecord
   has_many :article_categories, dependent: :destroy
   has_many :categories, through: :article_categories
   has_many_attached :attachments
+  has_many :feedbacks, dependent: :destroy
```
vishaltelangre: Is this relevant?

app/workers/daily_delete_temporary_article_worker.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class DailyDeleteTemporaryArticleWorker
```
vishaltelangre: TemporaryArticlesDailyCleanupWorker

app/workers/daily_delete_temporary_article_worker.rb
```diff
@@ -0,0 +1,16 @@
...
+  def perform
+    temporary_articles.destroy_all
+    logger.info("Temporary articles deleted successfully.")
+  end
+
+  private
+
+    def temporary_articles
+      Article.temporary_articles
+    end
```
vishaltelangre: Please extract this method into a stand-alone service. The benefit of a stand-alone service is the ability to unit test it in isolation without the need of a background job processing system in place.

app/workers/daily_delete_temporary_article_worker.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class DailyDeleteTemporaryArticleWorker
+  include Sidekiq::Worker
+
+  def perform
+    temporary_articles.destroy_all
```
vishaltelangre: Make a call to the extracted service here to delete the temporary articles whose age is about a couple of hours.

We do not want to delete a temporary article using the recurring rake task or cron job which is created just a minute ago.

vishaltelangre:
@tejaswinichile _a Please take a look at the comments. If there is a confusion about anything, we can have a call to discuss it in detail.

app/models/article.rb
vishaltelangre: But is this change relevant in this pull request?

Gemfile
vishaltelangre: @tejaswinichile 

1. Can you elaborate on why exactly we should move away from DelayedJob?
2. Have you checked whether Heroku supports background job processing using Sidekiq? Also, Sidekiq has a dependency on Redis, regarding which I am not sure whether Heroku provides a Redis dyno for free for the review apps or not.

cc: @neerajdotname

tejaswinichile: @vishaltelangre Yeah, Heroku does provide this support with `RedisToGo` free add-ons.

Sidekiq does provide this Heroku support and inbuilt Redis support. It is a plus point of having Redis to store the job instead of the application database. I tend to use sidekiq.

app/models/article.rb
tejaswinichile: Yes, else it's breaking while destroying the article records.

tejaswinichile:

tejaswinichile:
@vishaltelangre _a removed delayed job from project. please have a look.

Procfile
```diff
@@ -1,2 +1,3 @@
 web: bundle exec puma -C config/puma.rb
 release: bundle exec rake db:migrate && bundle exec rake perform_on_every_deploy
+worker: bundle exec sidekiq -C config/sidekiq.yml
```
vishaltelangre: Isn't this required in `Procfile.dev`?

app/services/temporary_articles_cleanup_service.rb
```diff
@@ -0,0 +1,18 @@
...
+class TemporaryArticlesCleanupService
+  def initialize
+    @articles = Article.temporary_articles
+  end
+
+  def process
+    article_saved_two_hours_ago.destroy_all
+  end
+
+
```
vishaltelangre: Remove additional blank line here.

app/workers/auto_close_resolved_tickets_worker.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class AutoCloseResolvedTicketsWorker
+  include Sidekiq::Worker
+
+  def perform
+    service = AutoCloseResolvedTicketService.new
+    service.process
+    logger.info("Auto closed resolved tickets.")
```
vishaltelangre: No need of the `logger.info` here, Sidekiq writes to its own log file. Also, the logged information has no actual usage.

app/workers/parse_user_agent_worker.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class ParseUserAgentWorker
+  include Sidekiq::Worker
+
+  def perform(ticket, user_agent)
+    if user_agent.present?
+      parsed_device_info = ParseUserAgentService.new(user_agent).parse
+      update(device_info: parsed_device_info)
```
vishaltelangre: Should it be `ticket.update_attributes!(device_info: parsed_device_info)`?

app/workers/parse_user_agent_worker.rb
```diff
@@ -0,0 +1,14 @@
...
+    else
+      update(device_info: nil)
+    end
```
vishaltelangre: Do we need this else part assuming that the Ticket#device_info is by default `nil`?

app/workers/temporary_articles_daily_cleanup_worker.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class TemporaryArticlesDailyCleanupWorker
+  include Sidekiq::Worker
+
+  def perform
+    TemporaryArticlesCleanupService.new.process
+    logger.info("Temporary articles deleted successfully.")
```
vishaltelangre: Remove the `logger.info` statement here since the logged information does not have any actual use-case.

vishaltelangre:
@tejaswinichile _a Please take care of the comments.

Procfile
tejaswinichile: I was doing it manually 👍

app/workers/auto_close_resolved_tickets_worker.rb
tejaswinichile: Yeah, In that log this line will get included.

app/workers/parse_user_agent_worker.rb
tejaswinichile: I need to check the default value here. Yeah we don't need this 👍

PR #672, opened 2018-10-04
Delete attachment from an article

tejaswinichile:
1. added route to destroy attachments
2. added attachment purge method to delete activestorage attachments
3. We are deleting single attachment at a time

### DIFF
--- a/app/controllers/articles/attachments_controller.rb
+++ b/app/controllers/articles/attachments_controller.rb
@@ -3,7 +3,7 @@
 class Articles::AttachmentsController < ApplicationController
   before_action :ensure_user_is_logged_in
   include LoadOrganization
-  before_action :load_article, only: :create
+  before_action :load_article, only: [:create, :destroy]
 
   def create
     attachments = @article.attachments.attach(
@@ -11,10 +11,19 @@ class Articles::AttachmentsController < ApplicationController
       content_type: attachment_params[:content_type],
       filename: attachment_params[:filename]
     )
-
-    response = { success: true, attachment_url: url_for(attachments.last) }
+    attachment = attachments.last
+    response = { success: true, attachment_url: url_for(attachment), attachment_id: attachment.id }
 
     render json: response, status: :created
+  end
+
+  def destroy
+    attachment = @article.attachments.find(params[:id])
+    if attachment.purge
+      render status: 200
+    else
+      logger.error attachment.errors.messages
+    end
   end
 
   private
--- a/app/controllers/articles/attachments_controller.rb
+++ b/app/controllers/articles/attachments_controller.rb
@@ -20,7 +20,7 @@ class Articles::AttachmentsController < ApplicationController
   def destroy
     attachment = @article.attachments.find(params[:id])
     if attachment.purge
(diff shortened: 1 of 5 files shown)


tejaswinichile:
@vishaltelangre: here is the path to delete the attachment with attachment_id
` DELETE: /organizations/:organization_api_key/articles/:article_id/attachments/:id`
I am checking if we can do it with attachment_url.
I have updated upload response with `attachment_id` though.

app/controllers/articles/attachments_controller.rb
```diff
@@ -3,20 +3,29 @@
...
-    response = { success: true, attachment_url: url_for(attachments.last) }
+    response = { success: true, attachment_url: url_for(attachments.last), attachment_id: attachments.last.id }
 
     render json: response, status: :created
   end
 
+  def destroy
+    attachment = @article.attachments.find(params[:id])
+    if attachment.purge
+      response = { status: :created, message: "Attachment deleted successfully." }
```
vishaltelangre: Please change the response status to `:ok` here.

vishaltelangre: No need to render a JSON response since it will not be used at all. Just the 200 status is sufficient for the delete action.

app/controllers/articles/attachments_controller.rb
```diff
@@ -3,20 +3,29 @@
...
+    else
+      response = { status: :internal_server_error, message: "Not able to delete attachment." }
```
vishaltelangre: If there is an actual error available, we can log it here for the debugging purposes. I think the custom error message is of no use to the frontend.

app/controllers/articles/attachments_controller.rb
```diff
@@ -3,20 +3,29 @@
...
 
   def create
     attachments = @article.attachments.attach(
       io: attachment_params[:file],
       content_type: attachment_params[:content_type],
       filename: attachment_params[:filename]
     )
-
-    response = { success: true, attachment_url: url_for(attachments.last) }
+    response = { success: true, attachment_url: url_for(attachments.last), attachment_id: attachments.last.id }
```
vishaltelangre: The `attachments.last` is repeated here, therefore it can be assigned to a variable above.

tejaswinichile:
@vishaltelangre _a You can check now, we needed to update the `nokogiri` gem because of vulnerability issue in the previous version

app/controllers/articles/attachments_controller.rb
```diff
@@ -3,20 +3,29 @@
...
 
     render json: response, status: :created
   end
 
+  def destroy
+    attachment = @article.attachments.find(params[:id])
+    if attachment.purge
+      render status: 200
+    else
+      logger.error attachment.errors.messages
```
vishaltelangre: @tejaswinichile _a Missing `render` statement in the else branch here.

vishaltelangre: Also, please add test coverage.

tejaswinichile:
@vishaltelangre _a

Issue #650, opened 2018-09-28
Url mapping to actual url to site


tejaswinichile:

tejaswinichile:
https://3.basecamp.com/3091787/buckets/9258656/todolists/1323569626

tejaswinichile:
- [ ]  Update URL new/edit for with columns: url_patter, url_rule, url_title. save this data to urls table
- [ ]  Update article page's urls selection with url_rule and url_pattern
- [ ]  fetch articles with url pattern matching when request comes from client widget
- [ ]  Update setup file with utl patter and rule.

tejaswinichile:
PR is up with all latest changes.

PR #678, opened 2018-10-05
Url mapping with client context

tejaswinichile:
Issue #650


tejaswinichile:
This PR is not updated one according to new changes we are having. Closing this one.

Issue #664, opened 2018-10-03
Confirm user about the pending actions while navigating away


vishaltelangre:
Pending actions could be any of the following.

1. Unsaved editor contents
2. A file being uploaded

Reference: https://3.basecamp.com/3091787/buckets/9258656/todos/1326760067

PR #680, opened 2018-10-05
Confirm user about the pending actions while navigating away

vishaltelangre:
Fixes https://github.com/bigbinary/acehelp/issues/664.

### DIFF
--- a/app/javascript/Admin/Page/Article/Create.elm
+++ b/app/javascript/Admin/Page/Article/Create.elm
@@ -31,6 +31,7 @@ import Html.Attributes exposing (..)
 import Html.Events exposing (..)
 import Page.Article.Common exposing (..)
 import Page.Errors exposing (..)
+import Page.View as MainView
 import PendingActions exposing (PendingActions)
 import Reader exposing (Reader)
 import Route exposing (..)
@@ -53,6 +54,7 @@ type alias Model =
     , attachmentsPath : String
     , originalArticle : Maybe Article
     , pendingActions : PendingActions
+    , showPendingActionsConfirmation : Acknowledgement Msg
     }
 
 
@@ -69,6 +71,7 @@ initModel =
     , attachmentsPath = ""
     , originalArticle = Nothing
     , pendingActions = PendingActions.empty
+    , showPendingActionsConfirmation = No
     }
 
 
@@ -100,6 +103,8 @@ type Msg
     | ChangeEditorHeight (Result Dom.Error Dom.Element)
     | ResizeWindow Int Int
     | AddAttachments
+    | IgnorePendingActions Msg
+    | HidePendingActionsConfirmationDialog
 
 
 update : Msg -> Model -> ( Model, List (ReaderCmd Msg) )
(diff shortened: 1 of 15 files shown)


vishaltelangre:
@ritesh404 _a Please review.

Here's the demonstration video of this feature - https://www.useloom.com/share/a1168c77539e40dbb17f28b656cd17e6.

cc: @neerajdotname

app/javascript/Admin/PendingActions.elm
```diff
@@ -0,0 +1,70 @@
...
+exists : String -> PendingActions -> Bool
+exists id pendingActions =
+    case pendingActions of
+        PendingActions actions ->
+            actions
+                |> List.any (\action -> action.id == id)
+
+
+remove : String -> PendingActions -> PendingActions
+remove =
```
ritesh404: `remove ` is just an alias for without? do we need this?

app/javascript/Admin/Page/Article/Create.elm
```diff
@@ -50,6 +52,9 @@ type alias Model =
     , errors : List String
     , success : Maybe String
     , attachmentsPath : String
+    , originalArticle : Maybe Article
+    , pendingActions : PendingActions
```
ritesh404: `pendingActions` is redundant as it is already present in the Main module and can be passed down to the update and view functions.

app/javascript/Admin/Page/Article/Edit.elm
```diff
@@ -181,13 +189,36 @@ update msg model =
...
+                            desc
+
+                    else
+                        PendingActions.empty
             in
-            ( { model | desc = newDesc, errors = errors }
+            ( { model
+                | desc = newDesc
+                , errors = errors
+                , pendingActions = newPendingActions
```
ritesh404: If we remove `pendingActions` from the submodule then it would mean we would have to intercept DescInput in main module also to update the `pendingActions` in main module. I am not sure if there is another/better way to do this.

app/javascript/Admin/PendingActions.elm
vishaltelangre: Yes, this is an alias function. The functionality of both the `remove` and the `without` functions is same. The difference is just about the semantics and usage. 

At some place, we need to remove a pending action from the pending actions list, for that purpose we use the `remove` function.

At some another place, we need to filter the pending actions without a particular pending list, for that purpose we use the `without` function.

app/javascript/Admin/Page/Article/Create.elm
```diff
@@ -216,6 +257,25 @@ update msg model =
...
             )
 
+        IgnorePendingActions nextMsg ->
+            let
+                nextCmd =
+                    Task.succeed ()
+                        |> Task.perform (always nextMsg)
+                        |> always
+                        |> Reader.Reader
+                        |> Strict
```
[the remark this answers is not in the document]
vishaltelangre: We will take care of this separately since the changes needed to do this will be irrelevant to this changeset.

vishaltelangre:
@ritesh404 _a I have taken care of the feedback comments related to removing the redundant `pendingActions` field from the models of the sub-modules. Now, we pass down the `pendingActions` as an argument to the `update` and `view` functions of the sub-modules whenever necessary. 

Please review again.

app/javascript/Admin/Page/Article/Create.elm
```diff
@@ -95,38 +101,55 @@ type Msg
...
         DescInput desc ->
-            ( { model | desc = Field.update model.desc desc }, [] )
+            let
+                newPendingActions =
+                    pendingActionsOnDescriptionChange
+                        pendingActions
+                        model.originalArticle
+                        desc
+            in
+            ( { model | desc = Field.update model.desc desc }, newPendingActions, [] )
```
ritesh404: This is the only place where the `pendingAction` is being updated. For now, I think this should be okay but I think we can still do a better job to avoid passing back the same pendingAction  back to Main module. We need to do this to keep the return type of the update method uniform across submodules
cc: @vishaltelangre

PR #683, opened 2018-10-05
Better error msg in case elastic search is not running

neerajdotname:
@vishaltelangre _a


neerajdotname:
Assume it is home brew and run with it.

PR #686, opened 2018-10-06
Fixed suggested article link

### LINKED ISSUE #665: Suggested article link is not working
When you click on the suggested article in client widget it's not opening article.

chiraggshah:
Fixes #665


ritesh404:
@chiraggshah code looks good. The default part just says that - run the update method in `SuggestedList` module and set the updated state. But we weren't doing anything in the update method of `SuggestedList` which is what led to the error as well.

PR #687, opened 2018-10-06
Search button improvements

### LINKED ISSUE #666: Client widget article search should work on click of enter
Search button is not visible on careforever site.
https://careforever-for-acehelp.herokuapp.com/pages/about

chiraggshah:
Fixes #666

### DIFF
--- a/app/javascript/Client/Section/Search/SearchBar.elm
+++ b/app/javascript/Client/Section/Search/SearchBar.elm
@@ -51,7 +51,7 @@ view model color =
             , placeholder "Search for an Article"
             ]
             []
-        , span [ onClick OnSearch, style "width" "40px" ] [ FontAwesome.search ]
+        , span [ onClick OnSearch ] [ FontAwesome.search ]
         ]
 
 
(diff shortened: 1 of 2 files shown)


app/javascript/Client/Section/Search/SearchBar.elm
```diff
@@ -44,14 +44,14 @@ type Msg
 
 view : Model -> String -> Html Msg
 view model color =
-    div [ class "ah-search-bar", style "background-color" color ]
+    Html.form [ class "ah-search-bar", style "background-color" color, onSubmit OnSearch ]
```
ritesh404: Why did we have to use a form here? `OnSearch` is anyway fired `onClick` of the span element
The issue here was to fix the styling. I think if you check the `search.scss` file you might find something that's breaking

chiraggshah:
@ritesh404 : I thought the issue #666 had 2 requirements.
- Search button is not visible on careforever site. (this was as per the description of the issue)
I handled this in the first commit

- Client widget article search should work on click of enter (this was the title of the issue)
I thought what this meant was that the search should work on press of enter too. Thats why I implemented Form and onSubmit here in the second commit.

If the second point is not needed, then I can revert the commit. But I think it would be good to have.

ritesh404:
@chiraggshah 

> Search button is not visible on careforever site. (this was as per the description of the issue)
I handled this in the first commit

Rather than inlining the style can you check why the styles specified we not working

> Client widget article search should work on click of enter (this was the title of the issue)
> I thought what this meant was that the search should work on press of enter too. Thats why I implemented Form and onSubmit here in the second commit.

Okay no worries. We can go with form submit

chiraggshah:
@ritesh404 _a : Padding was used instead of margin. Fixed it.

PR #692, opened 2018-10-08
 Map Category to URL

### LINKED ISSUE #689: Create graphQL mutation to map Categories to URL Patterns

ritesh404:
Fixes #689


ritesh404:
@tejaswinichile _a

tejaswinichile:
@ritesh404 _a I have added a few changes which we have discussed.
after that @vishaltelangre can you please review this PR. It's big it will take your time to check.

tejaswinichile:
@vishaltelangre _a You can review this PR.

app/controllers/application_controller.rb
```diff
@@ -5,7 +5,7 @@ class ApplicationController < ActionController::Base
 
   before_action :configure_permitted_parameters, if: :devise_controller?
   rescue_from ActiveRecord::RecordNotFound, with: :render_not_found
-  skip_before_action :verify_authenticity_token
+  protect_from_forgery with: :exception
```
vishaltelangre: Isn't https://github.com/bigbinary/acehelp/pull/694 going to take care of this?

app/graphql/mutations/url_mutations.rb
```diff
@@ -84,4 +89,38 @@ class Mutations::UrlMutations
...
+          else
+            errors = Utils::ErrorHandler.new.error("Categories not present.", context)
+          end
+        end
+      end
+      {
+        url: updated_url,
+        errors: errors
+      }
+    }
```
vishaltelangre: So many things are happening in this block. Can we extract it into smaller methods somehow? It is difficult to follow what's really going on here.

vishaltelangre: Actually, move this into a stand-alone service.

app/graphql/types/mutation_type.rb
```diff
@@ -36,6 +36,7 @@
 
   field :addUrlToArticle, field: Mutations::UrlArticleMutations::AddUrlToArticle.field
   field :removeUrlFromArticle, field: Mutations::UrlArticleMutations::RemoveUrlFromArticle.field
```
vishaltelangre: Since an URL is not associated with an article, do we still need `addUrlToArticle` and `removeUrlFromArticle` here?

app/graphql/types/mutation_type.rb
```diff
@@ -36,6 +36,7 @@
...
+  field :assignCategoryToUrlPattern, field: Mutations::UrlMutations::AssignCategoryToUrlPattern.field
```
vishaltelangre: Because a request may contain multiple categories, can we rename `assignCategoryToUrlPattern` to `assignCategoriesToUrlPattern`?

app/javascript/Admin/Admin/Data/Url.elm
```diff
@@ -77,15 +150,20 @@ urlByIdQuery =
 createUrlMutation : GQLBuilder.Document GQLBuilder.Mutation UrlResponse CreateUrlInput
 createUrlMutation =
     let
-        urlVar =
-            Var.required "url" .url Var.string
+        ruleVar =
+            Var.required "url_url" .url_rule Var.string
```
vishaltelangre: Is `url_url` a typo or intentionally named as such? It doesn't look good and should be renamed to something meaningful.

app/javascript/Admin/Helpers.elm
```diff
@@ -62,6 +65,23 @@ validUrl =
         |> Maybe.withDefault Regex.never
 
 
+validUrlRule : Regex
+validUrlRule =
+    Regex.fromStringWith { caseInsensitive = True, multiline = False }
+        "^[a-zA-Z0-9\\.\\/-_~!$&'\\(\\)\\*\\+,;=:@]+$"
+        |> Maybe.withDefault Regex.never
```
vishaltelangre: It will be good if we add a comment describing what all kinds of rules this regular expression can match. Otherwise, changing or extending it will be very tricky. I

 feel the need of a test case for this function to ensure that its behavior is consistent if a change is made to the regular expression.

cc: @neerajdotname

app/javascript/Admin/Page/Url/Edit.elm
```diff
@@ -120,76 +107,53 @@ update msg model =
...
+            case rule of
+                "is" ->
+                    ( { model | rule = Field.update model.rule <| updateRuleType UrlIs <| Field.value model.rule }, [] )
 
-        LoadSetting (Err err) ->
-            ( { model | errors = [ "Something went wrong while fetching the base Url" ] }, [] )
+                "contains" ->
+                    ( { model | rule = Field.update model.rule <| updateRuleType UrlContains <| Field.value model.rule }, [] )
+
+                "ends_with" ->
```
vishaltelangre: Instead of pattern-matching these strings such as `is`, `contains`, `ends_with` here, can we match it against concrete union type values? A small typo in the string can lead to a bug which is really hard to find.

app/javascript/Admin/Page/Url/Edit.elm
```diff
@@ -120,76 +107,53 @@ update msg model =
...
+view { errors, rule } =
+    commonView
+        { title = "Edit Url Pattern"
+        , errors = errors
+        , success = Nothing
+        , rule = rule
+        , onSaveUrl = UpdateUrl
+        , onUrlPatternInput = UrlPatternInput
+        , onRuleChange = RuleChange
+        , saveLabel = "Update Url"
```
vishaltelangre: Please change `Url` to `URL` at all places where we display it to the user. The function names or constant names can still have `Url`.

app/javascript/Admin/Page/Url/MapCategories.elm
```diff
@@ -0,0 +1,206 @@
...
+    , categories : List (Option Category)
+    }
+
+
+initModel : Model
+initModel =
+    { errors = []
+    , success = Nothing
+    , url = Nothing
+    , urlId = ""
```
vishaltelangre: It may happen that we have a model containing a value for the `url` field but the `urlId` is still blank, which will make the model to have an impossible state.

We probably should unify `urlId` and `url` fields into a union type at some point. For example.

```elm
type Url = NoUrl | UrlWithId UrlId | UrlWithIdAndData UrlId UrlData
```

app/models/article.rb
```diff
@@ -49,11 +47,63 @@ def self.search_using(org, options = {})
...
     if options[:search_string].present?
       articles = articles.search options[:search_string]
       articles = articles.each_with_object([]) { |article, arr| arr.push(article) }
     end
     articles
   end
+
+  private
+
+    def self.search_with_url_pattern(articles, incoming_url, org)
```
vishaltelangre: 1. Please expand `org` to `organization`.
2. I am not sure whether the class methods become private when defined beneath the  `private` directive. I think we need to use [`private_class_method`](http://ruby-doc.org/core-2.5.1/Module.html#method-i-private_class_method) to make it private. Actually, let's move all these class methods related to search into a stand-alone service. That way, we will not need to define class methods at all and the model will become skinny.

app/models/article.rb
```diff
@@ -49,11 +47,63 @@ def self.search_using(org, options = {})
...
+
+  # def self.pattern_matching_with_postgres_query(articles, url, org)
+  #   url_ids = Url.where("? ~* url_pattern", url).pluck(:id)
+  #   if url_ids.any?
+  #     articles = articles.joins(:urls).where(urls: { id: url_ids })
+  #   else
+  #     articles = []
+  #   end
+  #   articles
+  # end
```
vishaltelangre: Please don't leave a residue of the commented code.

app/models/article.rb
```diff
@@ -49,11 +47,63 @@ def self.search_using(org, options = {})
...
+      urls = org.urls.where(url_rule: :contains)
+      url_ids = []
+      urls.each do |url|
+        url_ids << url.id if incoming_url.include? url.url_pattern
+      end
+      url_ids
+    end
+
+    def self.urls_with_ends_with_rule(org, incoming_url)
+      urls = org.urls.where(url_rule: :ends_with)
```
vishaltelangre: Please define constants or enum types for the values of `url_rule` attribute.

vishaltelangre: It appears that we have an enum type for `url_rule` in the `Url` model. Can we fetch the value of the specified values using a constant instead of a symbol or a string?

app/models/url.rb
```diff
@@ -1,12 +1,31 @@
...
-  validates_uniqueness_of :url, case_sensitive: false
-  validates_with HttpUrlValidator
+  enum url_rule: {
+    contains: "contains",
+    ends_with: "ends_with",
+    is: "is"
+  }
 
+  # validates_uniqueness_of :url, case_sensitive: false
+  # validates_with HttpUrlValidator
```
vishaltelangre: Please remove the commented out code.

vishaltelangre:
@tejaswinichile _a Please take care of the comments left on the backend-side code. After that, @ritesh404 can take care of the comments left on the Elm code.

app/models/article.rb
tejaswinichile: What do you mean by `constant` here? The symbols we can use because of the enum declared.

vishaltelangre: I thought if there's a way to use constants such as `URL_RULE_ENDS_WITH` instead of a symbol `:ends_with`, then that would be better.

But it appears that no such thing exists. If there could've been constant available for the each enum, we could've used it. 

I suggested this initially because we may accidentally have `org.urls.where(url_rule: :end_with)` query. Probably this will silently run and yield no results since there is no such URL which matches url_rule having value `:end_with`. Note the difference between `:ends_with` and `:end_with` here.

Anyway, you can ignore the above comment since it cannot be done.

app/models/article.rb
tejaswinichile: Okay, we have article_search_service defined but we are not using it anywhere.
We were using it before for article API controller for widget article search.

tejaswinichile:
@vishaltelangre _a I have added rails side changes and some of the elm changes. Please check if you want me to improve anything on server side.

app/services/assign_categories_to_url_service.rb
```diff
@@ -0,0 +1,43 @@
...
+  def initialize(url_id, category_ids, organization)
+    @url_id = url_id
+    @category_ids = category_ids
+    @organization = organization
+  end
+
+  def process
+    @url = Url.find_by(id: url_id, organization_id: organization.id)
+    if url.nil?
+      @errors = Utils::ErrorHandler.new.error("Url not found", context)
```
vishaltelangre: Initialize `@errors = []` otherwise it will be set to `nil` by default.

app/services/assign_categories_to_url_service.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+class AssignCategoriesToUrlService
```
vishaltelangre: How about `CategoryToUrlAssignmentService` or `CategoryAndUrlMappingService`?

app/javascript/Admin/Page/Url/Edit.elm
ritesh404: @vishaltelangre `RuleChange` gets fired when the user selects an item in the drop down and the value returned is a string. We do this string pattern matching to actually convert this into a custom type `UrlRule`. I think there is no other way around this

vishaltelangre:
@tejaswinichile _a Added some minor comments. Please add the tests for the new services and then you can assign it to Ritesh to address the comments related to Elm code.

app/javascript/Admin/Page/Url/Edit.elm
vishaltelangre: Got it.

PR #694, opened 2018-10-09
Removed verify_auth_token skip

tejaswinichile:
1. Added csrf-xhr package to add csrf-token to header
2. csrf-xhr interrupt every xhr request and adds authenticity token to it.
3. updated application layout to include csrf-token
4. updated script template with available api key to render articles with authenticity token check


app/assets/javascripts/csrf-xhr.js
```diff
@@ -0,0 +1,61 @@
...
+                        ' because when csrf-xhr originally loaded, it could not find a <meta> element in the <head> with name="csrf-token" and a Rails CSRF token in a content attribute.'
+                );
+            } else {
+                this.setRequestHeader("X-CSRF-Token", csrfToken);
+            }
+        }
+
+        return originalResult;
+    };
+})(XMLHttpRequest.prototype);
```
prathamesh-sonpatki: @tejaswinichile Can we use the package via webpacker?

tejaswinichile: Let me check.

app/javascript/packs/script-template.js
```diff
@@ -1,6 +1,6 @@
 var req = new XMLHttpRequest();
 var baseUrl = "";
-var apiKey = "8272115d318bad004070";
+var apiKey = "9099015ee520e11887eb";
```
vishaltelangre: Can we pass this key to this file while compiling this script on the server by renaming the script as `script-template.js.erb` or something similar? Hard-coding this value can be very problematic and difficult to maintain.

tejaswinichile: This is only for ace-help client pages. This is the temporary one, otherwise, we were sending it through server only.
Removing this change for now.

tejaswinichile:

vishaltelangre:
@tejaswinichile _a

app/javascript/packs/client.js
```diff
@@ -1,4 +1,5 @@
 import { Elm } from "../Client/Main";
+import "csrf-xhr";
```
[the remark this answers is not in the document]
tejaswinichile: @vishaltelangre From mata_tags 

You can check it here : https://github.com/bigbinary/acehelp/blob/master/app/views/layouts/application.html.slim#L4

Reference: https://api.rubyonrails.org/classes/ActionView/Helpers/CsrfHelper.html

We had this from the beginning but we were not embedding it inside the header, which now we are achieving with this package.

vishaltelangre:
Since the widget can be embedded on any third party website which will not
have the CSRF token required by the AceHelp Rails app. Due to this,
Requests issued by the widget can not have a CSRF token but some sort of
auth token.

tejaswinichile:
@vishaltelangre I have added this in basecamp post: https://3.basecamp.com/3091787/buckets/9258656/todos/1331971485. Now we are separating the client widget request in controller level where we can add skip_verify_auth token, check https://github.com/bigbinary/acehelp/pull/703. For client widget request we don't need to check authenticity_token against the same token stored in the session cookie, as we don't have the session for the same.

app/controllers/graphql_api_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+    end
+
+    def request_is_mutation?
+      params[:query].starts_with?("mutation")
+    end
+
+    def request_is_mutation_for?(mutation_type)
+      request_is_mutation? && params[:query] =~ /#{mutation_type}/
+    end
+end
```
vishaltelangre: @tejaswinichile _a I do not like the name of this Controller and its endpoint. Since almost all functionality is duplicated in this new controller from the GraphqlController, can we reuse it somehow?

Let's put all the common stuff in `Graphql::BaseController`.
Then extend two sub-classes such as `Graphql::DashboardController` (having endpoint `/graphql/dashboard`) and `Graphql::WidgetController` (having endpoint `/graphql/widget`) from that base class respectively.

config/routes.rb
```diff
@@ -4,6 +4,7 @@
   root to: "home#index"
 
   post "/graphql", to: "graphql#execute"
+  post "/graphql_api", to: "graphql_api#execute"
```
vishaltelangre: Let's add a `resources` directive for this and rename `#execute` method to standard `#create` method.

vishaltelangre: We can think of alternate names such as `GraphqlExecutions::BaseController`, `GraphqlExecutions::WidgetController` and `GraphqlExecutions::DashboardController`. A **GraphQL (Query) Execution** can be created, therefore we can have a `#create` action in it.

tejaswinichile: Hmm...That makes sense. 👍

tejaswinichile:

app/controllers/graphql_execution/widgets_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+  skip_before_action :verify_authenticity_token
+
+  def create
+    result = AcehelpSchema.execute(query, variables: variables, context: context, operation_name: operation_name)
+    render json: result
+  rescue => e
+    show_error_in_logs(e)
+    graphql_error = Utils::ErrorHandler.new.generate_graphql_error_with_root(e.message, path: ["System Exception"])
+    render json: graphql_error, status: 500
+  end
```
vishaltelangre: Cannot we put this #create action into the base controller? 

I do not see any difference between the #create action of GraphqlExecution::WidgetsController and GraphqlExecution::DashboardsController.

app/controllers/graphql_execution/dashboards_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+  private
+
+    def context
+      context = {}
+      context[:organization] = @organization if @organization.present?
+      context[:current_user] = current_user
+      context[:request] = request
+      context
+    end
+end
```
vishaltelangre: Same for this method, it can extracted out in the base controller.

It can be overridden here as

```rb
def context
  super.merge(current_user: current_user)
end
```

vishaltelangre:
@tejaswinichile _a Please take a look at my recent comments.

tejaswinichile:
@vishaltelangre _a

app/controllers/graphql_execution/dashboard_controller.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class GraphqlExecution::DashboardsController < GraphqlExecution::BaseController
+class GraphqlExecution::DashboardController < GraphqlExecution::BaseController
   include LoadOrganization
```
vishaltelangre: `include LoadOrganization` is already present in the base controller, please remove it.

app/controllers/graphql_execution/widget_controller.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+class GraphqlExecution::WidgetController < GraphqlExecution::BaseController
+  include LoadOrganization
```
vishaltelangre: Same here, please remove this include statement since it is already present in the base controller.

vishaltelangre:
@tejaswinichile _a Please take care of the latest comments and after addressing them, you can merge this pull request yourself.

PR #703, opened 2018-10-09
added graphql api controller

tejaswinichile:
1. updated client request helper
2. added graphql api endpoint for client widget request
3. updated script template with api-key


tejaswinichile:
Will be including this changes in #694

Issue #685, opened 2018-10-06
Delete temporary article, when user move away from create article page.


tejaswinichile:
When user navigate away from article create page without saving the data.
Delete the article along with the attachments.

vishaltelangre:
@tejaswinichile If you have not started work on this, can I take this up?

tejaswinichile:
Yeah. I haven't. You can take this up.

vishaltelangre:
@vishaltelangre _a

PR #709, opened 2018-10-10
Delete temporary article when user clicks the cancel button in create article page

vishaltelangre:
Fixes https://github.com/bigbinary/acehelp/issues/685.

### DIFF
--- a/app/javascript/Admin/Page/Article/Create.elm
+++ b/app/javascript/Admin/Page/Article/Create.elm
@@ -103,6 +103,8 @@ type Msg
     | AddAttachments
     | IgnorePendingActions Msg
     | HidePendingActionsConfirmationDialog
+    | DiscardArticle
+    | DeleteTemporaryArticleResponse
 
 
 update : Msg -> PendingActions -> Model -> ( Model, PendingActions, List (ReaderCmd Msg) )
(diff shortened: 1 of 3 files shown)


vishaltelangre:
@tejaswinichile _a Please review.

app/javascript/Admin/Page/Article/Create.elm
```diff
@@ -253,6 +255,25 @@ update msg pendingActions model =
         HidePendingActionsConfirmationDialog ->
             ( { model | showPendingActionsConfirmation = No }, pendingActions, [] )
 
+        DiscardArticle ->
+            if PendingActions.isEmpty pendingActions then
```
tejaswinichile: What does `pendingActions` do?

vishaltelangre: @tejaswinichile PendingActions was introduced in https://github.com/bigbinary/acehelp/pull/680. It keeps a list of ongoing actions which are not finished yet. When discarding an article, we check whether there are any pending actions or not. There may be a pending action for a ongoing file upload or a pending action for the unsaved contents in the editor. If there exists at least a pending action, we display a confirmation dialog to the user describing that there are some pending actions which is preventing the discard article action. The user may choose to ignore the pending actions and still discard the article.

Issue #711, opened 2018-10-10
Not able to copy the widget's embed code in Firefox browser


vishaltelangre:
https://www.useloom.com/share/0d847d43186f4cc89a78d3f1e86607fd

PR #712, opened 2018-10-10
Fix: Not able to copy the widget's embed code in Firefox browser

vishaltelangre:
Fixes https://github.com/bigbinary/acehelp/issues/711.


tejaswinichile:
Tested on review app, this change not affecting to chrome or safari.

PR #714, opened 2018-10-10
Edit category changes

chiraggshah:


chiraggshah:
@ritesh404 : This PR is not complete yet. I wanted to refresh the page if the category is edited successfully, I tried but was not able to figure out. Can you help or push me in the right direction on how to do that.
Apart from that, there's some code cleanup that needs to be done which I will do after page refresh and deliver this PR. Thanks

PR #715, opened 2018-10-10
Redirect to article list page on organization change

### LINKED ISSUE #688: On changing organization navigate the user back to the Article tab
Currently, if the user changes the organization. The organization changes but the user still remains in the current page. This causes a bug. 

Consider the following scenario 
- User logs in
- User goes to Article tab and selects an Article
- User Changes the organization from hamburger menu
- Notice that the organization changes but the user is presented with an error message saying that the article does not exist. This happens because the user is in the article show page with an article id of the previous organization. This Id does not exist for the current organization

So as a solution to this we can navigate the user back to Article listing page by default on changing the organization

chiraggshah:
Fixes #688

### DIFF
--- a/app/javascript/Admin/Main.elm
+++ b/app/javascript/Admin/Main.elm
@@ -1167,7 +1167,7 @@ view model =
                     [ viewWithTopMenu
                     , MainView.hamBurgerMenu
                         { organizationList = model.organizationList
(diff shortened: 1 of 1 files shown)


chiraggshah:
@ritesh404 _a : Please review

vishaltelangre:
Looks good to me.

PR #720, opened 2018-10-11
Updated redis pool and concurrency size

tejaswinichile:
Redis concurrency issue

### DIFF
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -1,7 +1,7 @@
 Sidekiq::Extensions.enable_delay!
 
 Sidekiq.configure_server do |config|
-  config.redis = { url: ENV['REDIS_URL'] }
+  config.redis = { url: ENV['REDIS_URL'], size: 3 }
   schedule_file = Rails.application.secrets.sidekiq_schedule_file
 
   if File.exists?(schedule_file)
@@ -10,5 +10,5 @@ Sidekiq.configure_server do |config|
 end
 
 Sidekiq.configure_client do |config|
-  config.redis = { url: ENV['REDIS_URL'] }
+  config.redis = { url: ENV['REDIS_URL'], size: 1 }
 end
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -1,7 +1,7 @@
 Sidekiq::Extensions.enable_delay!
 
 Sidekiq.configure_server do |config|
-  config.redis = { url: ENV['REDIS_URL'], size: 3 }
+  config.redis = { url: ENV['REDIS_URL'], size: 9 }
   schedule_file = Rails.application.secrets.sidekiq_schedule_file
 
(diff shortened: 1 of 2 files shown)


tejaswinichile:
Reference:  http://manuelvanrijn.nl/sidekiq-heroku-redis-calc/

tejaswinichile:
@vishaltelangre _a Can you check. We are not yet sure about this solution but we will try it out.

config/sidekiq.yml
```diff
@@ -1,5 +1,5 @@
 default: &default
-  :concurrency: 5
+  :concurrency: 1
```
vishaltelangre: Can we set this to max concurrency allowed by RedisToGo addon?

config/initializers/sidekiq.rb
```diff
@@ -1,7 +1,7 @@
 Sidekiq::Extensions.enable_delay!
 
 Sidekiq.configure_server do |config|
-  config.redis = { url: ENV['REDIS_URL'] }
+  config.redis = { url: ENV['REDIS_URL'], size: 3 }
```
vishaltelangre: According to https://github.com/mperham/sidekiq/wiki/Using-Redis#complete-control, the Sidekiq server requires connections of size "concurrency + 5".

vishaltelangre:
@tejaswinichile _a

config/initializers/sidekiq.rb
tejaswinichile: Yeah, but we need to take care of the number of workers we have.

config/sidekiq.yml
tejaswinichile: RedisToGo provide max 10 connections. According to that we have to define sidekiq concurrency.

config/initializers/sidekiq.rb
tejaswinichile: This is what redisToGo provide as max 10

tejaswinichile:
Let me set it at max size

tejaswinichile:
@vishaltelangre Can you check now I have just increased Redis connection size, kept concurrency same as before.

config/initializers/sidekiq.rb
```diff
@@ -1,7 +1,7 @@
 Sidekiq::Extensions.enable_delay!
 
 Sidekiq.configure_server do |config|
-  config.redis = { url: ENV['REDIS_URL'], size: 3 }
+  config.redis = { url: ENV['REDIS_URL'], size: 9 }
```
vishaltelangre: @tejaswinichile _a If this value is available using some environment variable, then please use that environment variable instead of this hard-coded value here. 

Otherwise, it's okay, just add the link for reference about this number in that case.

Once this has been addressed, please merge the pull-request yourself.

tejaswinichile: @vishaltelangre I have added the link to reference, we will have secret or constant declared for the same Once we are sure that this is the proper solution for this issue.

Issue #713, opened 2018-10-10
Check if the widget is actually installed and usable


vishaltelangre:
Until the widget is not accessed at all, we can display that the widget is not installed in the "Settings" page.

PR #721, opened 2018-10-11
Check if the widget is actually installed and usable

vishaltelangre:
Fixes https://github.com/bigbinary/acehelp/issues/713.


vishaltelangre:
@tejaswinichile _a Please review.

app/controllers/graphql_controller.rb
```diff
@@ -71,4 +74,8 @@ def request_is_mutation?
     def request_is_mutation_for?(mutation_type)
       request_is_mutation? && params[:query] =~ /#{mutation_type}/
     end
+
+    def requesting_client_is_widget?
+      request.headers["X-Client"] == WIDGET_CLIENT_REQUEST_HEADER
```
tejaswinichile: 1. As we will be having the separate controller, this change would be irrelevant.
You can directly set the https://github.com/bigbinary/acehelp/pull/721/files#diff-e54f3c67ff47c764363e39152508a4ddR36 in context if the request is hitting to graphql widget controller.
No need to maintain the var in headers.

app/graphql/resolvers/organizations_search.rb
```diff
@@ -3,7 +3,14 @@
 class Resolvers::OrganizationsSearch < GraphQL::Function
   type Types::OrganizationType
 
+
   def call(obj, args, context)
-    Organization.find_by(api_key: context[:organization].api_key)
+    organization = Organization.includes(:setting).find_by(api_key: context[:organization].api_key)
+
+    if organization && context[:requesting_client_is_widget]
```
tejaswinichile: Instead of having such check we can have separate organization search resolver for those requests which are coming through client widget. We can have common things at one place just an extra update call will be in widget's organization search.
Your thoughts?

vishaltelangre:
Thanks, @tejaswinichile for the feedback. 

Since there was no way to differentiate between the `POST /graphql` requests issued by the widget and by the dashboard site, I had to include the request `X-Client` request header to overcome that issue.

With https://github.com/bigbinary/acehelp/pull/694, reliance on that `X-Client` header won't be required due to the introduction of the separate controllers to handle the incoming GraphQL requests issued by the both widget and the dashboard site respectively.

I will make the necessary changes once the pull request https://github.com/bigbinary/acehelp/pull/694 is merged.

tejaswinichile:
@vishaltelangre _a Can make respective changes to this PR. According to #694

vishaltelangre:
@tejaswinichile _a Please review.

Issue #707, opened 2018-10-09
Remove ruby-audit cve skip


vipulnsward:
Remove this skip:

--ignore CVE-2016-10545

PR #722, opened 2018-10-11
Do not ignore check for CVE-2016-10545 since it is not a security issue

vishaltelangre:
The CVE with ID CVE-2016-10545 has been rejected.
According to https://nvd.nist.gov/vuln/detail/CVE-2016-10545, investigation showed that it was not a security issue.

Fixes https://github.com/bigbinary/acehelp/issues/707.


vishaltelangre:
@vipulnsward _a Please review. Build on Circle CI is passing.

PR #725, opened 2018-10-11
Show Comment Info

### LINKED ISSUE #718: Posted three different replies
https://cl.ly/38ee04b4f3f1

For each reply post the name of the person who posted reply and the date. Do not worry about format and UI right now. We will fix that later. Right now let's get the data right.

@chiraggshah _a

chiraggshah:
Fixes #718

### DIFF
--- a/app/javascript/Admin/Page/Ticket/Edit.elm
+++ b/app/javascript/Admin/Page/Ticket/Edit.elm
@@ -64,7 +64,7 @@ initModel ticketId =
     , statuses = []
     , comments = []
     , notes = []
-    , comment = Comment ticketId "" ""
+    , comment = Comment ticketId "" "" ""
     , agents = []
     , agent = Nothing
     }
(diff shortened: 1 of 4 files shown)


chiraggshah:
@ritesh404 _a : Please review

cc @neerajdotname

app/javascript/Admin/Page/Ticket/Edit.elm
```diff
@@ -64,7 +64,7 @@ initModel ticketId =
     , statuses = []
     , comments = []
     , notes = []
-    , comment = Comment ticketId ""
+    , comment = Comment ticketId "" "" ""
```
ritesh404: @chiraggshah can you change this to 

```
{   ticket_id  = ticketId
    , info = ""
    , created_at : ""
    , user_name : ""
}
```

In this case, I feel we should stick to creating records this way because it makes the code a little more clear. Record constructors can be used when there are 2 or at most 3 fields in a record otherwise, we have to check the type definition to see what position the fields are in.
Please do this in other places too

ritesh404:
@chiraggshah _a

Issue #732, opened 2018-10-12
Update timeoutable time for devise


tejaswinichile:
Setting timeoutable to 1.week for the app.

PR #733, opened 2018-10-12
updated devise configuration for session timeout

tejaswinichile:
Issue #732


tejaswinichile:
@vishaltelangre Merging this as not a big change here.

PR #734, opened 2018-10-12
Added command to start redis-server for worker

chiraggshah:
I updated master and was trying to run `foreman start -f Procfile.dev`, but it was showing errors.
I saw that we had moved from delayed job to sidekiq and the error was because it could not connect to redis.

Instead of having to run `redis-server` to start redis in another terminal tab, should we include it in the procfile?

@neerajdotname @tejaswinichile @ritesh404 @vishaltelangre  thoughts?

### DIFF
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -1,3 +1,3 @@
 web: bundle exec rails s -p 3000
 webpacker: ./bin/webpack-dev-server
-worker: bundle exec sidekiq -C config/sidekiq.yml
+worker: redis-server && bundle exec sidekiq -C config/sidekiq.yml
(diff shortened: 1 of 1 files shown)


Procfile.dev
```diff
@@ -1,3 +1,3 @@
 web: bundle exec rails s -p 3000
 webpacker: ./bin/webpack-dev-server
-worker: bundle exec sidekiq -C config/sidekiq.yml
+worker: redis-server && bundle exec sidekiq -C config/sidekiq.yml
```
vishaltelangre: @chiraggshah _a I assume that the `redis-server` command here would start in the foreground mode and therefore it would block the main thread in the current TTY session. Due to this, the `bundle exec sidekiq` command would never start.

Can you please check whether this is the case or not?

If we could start the `redis-server` command in the daemon mode (or background mode), then we can merge it. Otherwise, we can enlist the `redis-server` command as another process in the `Procfile.dev` file.

chiraggshah:
@vishaltelangre : You are right. I changed it now. Please check

Issue #729, opened 2018-10-11
Priority for tickets


neerajdotname:
Tickets will have - low, medium, high and urgent priority. Default value is medium.

Show this value in the list https://monosnap.com/file/zXpwAWk2WIRMJjwusjjMGgnd1XjFv1

@chiraggshah _a

PR #742, opened 2018-10-12
Add Priority to Tickets

chiraggshah:
Fixes https://github.com/bigbinary/acehelp/issues/729

image


chiraggshah:
@neerajdotname _a : Please review. This PR only implements adding the column and showing the value on the list page. I think next issue should be to show the priority and change the priority from the list page?

